PR #6, opened 2016-09-27
Add missing .mtz and include .log file into mtz/pdb download archive

isikhar:
Fixed duplicated .pdb file reference and included .log fine into archive for improved usability.


stufisher:
good spot

PR #10, opened 2017-03-01
Use modules system to find supported software versions

Anthchirp:
Suggested changes to use the modules system once again.
This will need some testing by @stufisher since we don't have any access to the user this is running under.
If it doesn't work then we should get it fixed from IT side.

### DIFF
--- a/api/includes/pages/class.mc.php
+++ b/api/includes/pages/class.mc.php
@@ -208,9 +208,7 @@
                 }
 
                 # /4479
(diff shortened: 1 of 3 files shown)


graeme-winter:
Did this get merged?

Anthchirp:
We have now removed all 6.4 and 6.5 ccp4s. Let us know if you are still referencing any of those if we need to restore them.

stufisher:
This doesnt work as i expected. If you have time to really dig into this please do, if not im updating the paths to newer version...

PR #18, opened 2017-08-16
SCI-6027 Add support for fast_ep html report

isikhar:
If available, show HTML report of fast_ep downstream processing task instead of plain-text log file.


stufisher:
Have a file list for fasp_ep instead now

PR #49, opened 2019-07-30
Drop execute bit on non-executable files

jlmuir:


drnasmith:
The build system changes (webpack/composer) mean most of these files are not required anymore. However I have updated the files which should be fixed in a separate branch (fix/executable_bit_images).

PR #58, opened 2019-11-04
add spine and actor puck options

stufisher:


drnasmith:
Currently deployed on release branch at DLS

PR #60, opened 2019-11-05
Fix typos

jlmuir:


stufisher:
The country spellings are not typos, these adhere to DHLs internal country list and must be kept as such

jlmuir:
Oh brother.  What are they thinking?!  "United States Of America"?  Ridiculous.  On top of that, some have a lowercase "of" and others an uppercase "Of".

It would be nice to decouple SynchWeb from the DHL country list.  What happens if support is added for FedEx or UPS?

Anyway, I've removed that commit and force-pushed.  The other two commits remain.

PR #67, opened 2020-01-02
sci-8960 proposal state

stufisher:
disable contact, shipment, dewar, container, and protein create if proposal is not active


stufisher:
need to rebase this off migrate/integration-rc1...

PR #83, opened 2020-03-26
SCI-7442: Update big_ep pannel to include all downstream processing

isikhar:
Updated BigEP panel that reads information from ProcessingJob and AutoProc tables and displays snapshots of the built models.
This change requires replacement of BigEP entries in config.php with
`"Big EP" => array('big_ep/', '', '', 'big_ep')`


drnasmith:
On hold until we know the impact of changing the main SQL query for ap_statuses

stufisher:
Superseded by https://github.com/DiamondLightSource/SynchWeb/pull/193 (rebased on master)

PR #90, opened 2020-04-15
Local customisation

stufisher:
Allows some local customisation, changing site name, url, header image from config.json (and builds into webpack bundle)

Allow specifiying port and proxy location as --env on npm run server

(rebased off master)


stufisher:
I should prolly document, e.g.
`npm run serve -- --env.port=9005 --env.proxy.target=http://localhost:80/sw`

PR #91, opened 2020-04-18
dont exclude lb visits from next (some are not scheduled=0)

stufisher:


drnasmith:
Deployed on a release branch for this run

PR #92, opened 2020-04-21
Updated README to describe build process

drnasmith:


KarlLevik:
> New ISPyB web application which

New? It was new maybe 5 years ago :)

> Running SynchWeb requires setting up a Linux, Apache, MySQL and PHP (LAMP) software stack.

I would suggest specifically saying MariaDB rather than MySQL. MySQL and MariaDB are diverging more every day, and we may not want to try and support both going forward. MariaDB is also what is used by all participants in the collaboration, I believe.

PR #106, opened 2020-05-23
sci-8602 centralise dc view code and allow update in place

stufisher:
This has become a bit more involved than i expected. Trying to centralise a lot of duplicate dc code at the same time as allowing views to update themselves


stufisher:
If a value in the dc changes and there exists a selector with the same name it will be automatically updated. In addition you can specify a function on the view to transform the value somehow.

```
<span class="DCC"><%-DCC%></span>
```
as DCC updates it will update the value in the span

```
updateDCC(value) {
    var transform = something(value)
    return transform
}
```

PR #108, opened 2020-06-07
Generalise per image

stufisher:


client/src/js/modules/dc/models/distl.js
```diff
@@ -37,9 +37,8 @@ define(['backbone'], function(Backbone){
         var d = this.get('data')
         if (d && d[0].length > 0) {
           if (this.nimg == _.last(d[0])[0]) refresh = false
-          if (this.pm.get('AGE') > 15) refresh = false
         }
-
+        if (this.pm.get('AGE') > 15) refresh = false
```
drnasmith, line 41: Some comments would be useful here.
Does pm mean parent model, if so what does that refer to in this context?
Also what is the AGE in this case, seconds since last DC update?

PR #110, opened 2020-06-08
feature/i03-374

drnasmith:
Adds the ability to clone proteins if the protein has a valid user office id.
Reuses Add Protein view.
Also takes away ability to edit the name - as requested in ticket for valid samples only; so users can easily associate clones/variants with the same protein name.
Preserves PDB associations.


stufisher:
Urgh, forgot to click submit review :|

PR #113, opened 2020-06-24
SCI-7989: sample user folder

stufisher:
WIP. Also resolves I03-390


drnasmith:
This update will run on staging until GDA can make use of the userPath

stufisher:
Need to actually test the api first, was waiting for db change ;)

PR #117, opened 2020-06-24
allow setting timezone / remove hardcoded dls reference

stufisher:


stufisher:
Where did i miss it? index.php now exports it?

PR #119, opened 2020-07-10
Consistent use of cas certificate in curl requests

drnasmith:
Explicitly set the cas certificate when making curl requests.
Should be more flexible for other sites using different CAS providers.


stufisher:
Cant see any problems with this, as long as its tested with vmxi ;) What does this fix?

drnasmith:
At Diamond its not required. However, when I generated independent certificates for a test CAS server and SynchWeb with the development environment (https://github.com/DiamondLightSource/synchweb-devel-env), its needed in that case. So it would be required for any site that uses separate certificates for CAS and SynchWeb installations.

PR #123, opened 2020-07-28
Client build fixes

jlmuir:


stufisher:
I dont think you should need to do this i need to check. Webpack uses the homepage variable to tell it if the root build is not /?

jlmuir:
As far as I can tell, I need it.  But yes, please check since I know next to nothing about webpack.  What do you mean by "homepage variable"?  I don't know what that is.  Am I supposed to set it somewhere?  FWIW, I tried with `apiurl` and `appurl` set to `/synchweb/api` and `/synchweb`, respectively, in `config.json`, and it didn't work (e.g., paths in `index.php` were wrong).

stufisher:
No you are absolutely right, these are needed. I've been working a lot with create-react-app recently and that has a "homepage": "/synchweb" variable you can add to package.json. Sorry for confusion!

drnasmith:
If we move to the vue-router approach (a fork in preparation)  it has a similar 'base' setting which does the same thing for the client.
Remember if you have a different apiurl the apache or nginx config needs to match.

drnasmith:
Plan to close this as webpack has been removed from titles.
Also base app url will be set via vue-router in latest vue master.

morybc:
Closing off, to tidy up codebase - this is over two years old and does not appear to be needed.  Please reopen with clear explanation of what it is supposed to achieve, if the change is desired to be kept.

PR #125, opened 2020-07-30
url redirects for trailing slash and uppercase urls

stufisher:
This will remove any trailing slash and convert any uppercase characters to lowercase. I dont think we have any urls in synchweb with casing but need to check. Also if other sites populate visits in uppercase this may break things. Hopefully everyone uses lower case...

### DIFF
--- a/client/src/js/app.js
+++ b/client/src/js/app.js
@@ -67,6 +67,15 @@ function(Backbone, Marionette, _, $, HeaderView, SideBarView, DialogRegion, Logi
 
   app.parseQuery()
 
+  // Strip trailing forward slash sci-8800
+  if (location.href.endsWith('/')) {
+    location.href = location.href.replace(/\/+$/, '')
+  }
+
+  // Convert uppercase url to lowercase sci-8799
+  if (location.href.match(/[A-Z]/)) {
+    location.href = location.href.toLowerCase()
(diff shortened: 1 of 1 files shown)


client/src/js/app.js
```diff
@@ -67,6 +67,15 @@ function(Backbone, Marionette, _, $, HeaderView, SideBarView, DialogRegion, Logi
...
 
+  // Strip trailing forward slash sci-8800
+  if (location.href.endsWith('/')) {
+    location.href = location.href.replace(/\/+$/, '')
+  }
+
+  // Convert uppercase url to lowercase sci-8799
+  if (location.href.match(/[A-Z]/)) {
+    location.href = location.href.toLowerCase()
+  }
```
Anthchirp, line 78: Why not run these unconditionally?
```suggestion
  // Strip trailing forward slashes SCI-8800 and convert to lower case SCI-8799
  location.href = rtrim(location.href.toLowerCase(), "/")
```

stufisher: infinite redirect :)

Anthchirp: Fair 🙂

drnasmith:
Closing this because we are using vue router now. It deals with these cases for us.

PR #135, opened 2020-08-20
Support client zero-padding in sample name clone

jlmuir:
If a sample name ends in a number that is zero-padded, preserve zero-padding when cloning it in the client.

For example, in the Shipments > [Shipment] > Add New Container view, when you click the Clone from First Sample button, if the name of the first sample is `Sample-01`, the generated sample names will be `Sample-02`, `Sample-03`, etc. instead of `Sample-2`, `Sample-3`, etc.


drnasmith:
Will perform some tidy up with the naming and comments post merge

PR #139, opened 2020-08-25
Fix Proposals breadcrumb: s/proposals/proposal/

jlmuir:
The Proposals breadcrumb link is broken since it links to `/proposals` when it should be `/proposal` (singular).


drnasmith:
Actually I think its the route that should be fixed rather than the breadcrumb. If the vue-router change makes it this run it will be addressed there. If not I'll fix it using the old controller/router files

PR #140, opened 2020-08-29
make valid_components server side

stufisher:
needs some more testing


stufisher:
renamed it to valid_components as its the proteins that have to be valid not the samples

PR #145, opened 2020-09-07
Change client page title from ISPyB to SynchWeb

jlmuir:
While it's true that the underlying database schema is ISPyB, what the user really sees and uses is SynchWeb.  The API is a SynchWeb API, so SynchWeb is what a developer sees and uses as well.  To make this more clear, change ISPyB in the client page title to SynchWeb.

This may be a controversial change.  I think it's a good change to make, but I would understand if the maintainers do not agree.  If this PR is rejected, would you be willing, just for my own understanding, to very briefly explain the rationale behind keeping the ISPyB naming?  Thanks!

### DIFF
--- a/client/src/js/views/header.js
+++ b/client/src/js/views/header.js
@@ -88,7 +88,7 @@ define(['marionette', 'backbone', 'templates/header.html'], function(Marionette,
         
         setTitle: function() {
             console.log('set title')
-            document.title = 'ISPyB » ' + this.bc.pluck('title').join(' » ')
+            document.title = 'SynchWeb » ' + this.bc.pluck('title').join(' » ')
         },
         
     })


drnasmith:
For users, synchweb is their only view into ISPyB.

In future the web application could change but the data and URL they interact with would still be ISPyB. For us that's ispyb.diamond.ac.uk but presume it could be ispyb.organisation.something. That gives sites flexibility for how they present the data to users over a long term while making it consistent for users.

The new vue version of synchweb includes a splash screen which makes the distinction clearer but highlights that its an interface/view into ISPyB metadata.

KarlLevik:
Only speaking for myself here, but wouldn't it be ideal if the page title were configurable? Maybe with a default value of 'ISPyB'?

I think some of our users would be utterly confused if they had to make a distinction between web application and database. So to them it's usually all just "ispyb" even though many of them are at least aware there are multiple components involved.

jlmuir:
> Only speaking for myself here, but wouldn't it be ideal if the page title were configurable? Maybe with a default value of 'ISPyB'?

Yes, I'd be fine with it being configurable.

> I think some of our users would be utterly confused if they had to make a distinction between web application and database. So to them it's usually all just "ispyb" even though many of them are at least aware there are multiple components involved.

Yes, but what I'm suggesting is that what your users know is actually SynchWeb, not ISPyB, so it would make more sense to brand it as SynchWeb.  You can tell this by considering what would happen if you replaced SynchWeb with EXI or ISPyB UI.  I predict your users would be utterly confused.  However, if you kept SynchWeb but replaced the ISPyB DB schema with something else, your users would have no idea that you'd made such a change and would continue to happily use SynchWeb as they always have.  Therefore, I conclude that your users know SynchWeb, not ISPyB, so it would make more sense to call it SynchWeb.  Anyway, if it's configurable, we can both easily call it what we want.

morybc:
This change also needs to be made to the breadcrumbs.vue file to work in the newer code.  I think I agree with @jlmuir but because this is a source of some disagreement I think it's best to just leave it as it is.

PR #146, opened 2020-09-07
sci-9302: make image tests and serving consistent

stufisher:
Need to check this is still performant as it does a few extra queries for the template parser

### DIFF
--- a/api/src/Page/DC.php
+++ b/api/src/Page/DC.php
@@ -628,8 +628,6 @@ class DC extends Page
             foreach ($dct as $d) $dcs[$d['ID']] = $d;
             
             $out = array();
-            
-            $tmp = new TemplateParser($this->db);
             foreach ($dcs as $dc) {
                 $debug = array();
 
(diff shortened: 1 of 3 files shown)


stufisher:
@dsclassen

stufisher:
Are the config variables correctly setup?

```
    # Visit directory on disk
    $visit_directory = '/dls/<%=BEAMLINENAME%>/data/<%=YEAR%>/<%=VISIT%>';

    # Diffraction image snapshots
    $jpeg_location = '<%=VISITDIR%>/jpegs/<%=IMDIRECTORY%>/<%=IMFILE%>.jpeg';
    $jpeg_thumb_location = '<%=VISITDIR%>/jpegs/<%=IMDIRECTORY%>/<%=IMFILE%>.thumb.jpeg';
```

KarlLevik:
Not sure if relevant, but note that Diamond has visit directories as both `/dls/$beamline/data/$year/$visit` and `/dls/mx/data/$proposal/$visit` .

stufisher:
Yes and because we no longer have the image table, we have to pick to support only one.

drnasmith:
Yes the config.php property is the same as that set in the config_sample.php file.

stufisher:
Urgh was caching the visit dir which was wrong

PR #153, opened 2020-09-29
add option to proxy assets directory

stufisher:


drnasmith:
Following the discsussion I have reverted the assets proxy target - see PR #158

PR #157, opened 2020-10-02
Grid view: check box is rendered before setting inner text

stufisher:


drnasmith (commented):
Not sure what issue this fixes. Doesn't seem to affect grid scans that I can see. Which checkbox does it refer to?

stufisher:
If you check in the console on every grid plot there is an error...

drnasmith:
OK - does not seem to manifest in any obvious omission but good to fix!

PR #169, opened 2020-10-26
Fix: Change how beam values are displayed

vonvick:
#### What does this PR do?
This PR changes how refined x and y beam values are displayed by formatting only values that are numeric.

### DIFF
--- a/client/src/js/templates/dc/dc_autoproc.html
+++ b/client/src/js/templates/dc/dc_autoproc.html
@@ -43,27 +43,19 @@
         <tbody>
             <tr>
                 <td>Start</td>
-                <td><% print(BEAM.XBEAM ? BEAM.XBEAM : '- -') %></td>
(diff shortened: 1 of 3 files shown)


client/src/js/templates/dc/dc_autoproc.html
```diff
@@ -43,19 +43,19 @@
         <tbody>
             <tr>
                 <td>Start</td>
-                <td><%-BEAM.XBEAM%></td>
-                <td><%-BEAM.YBEAM%></td>
+                <td><%= BEAM.XBEAM ? BEAM.XBEAM : '- -' %></td>
```
stufisher, line 46: Make sure to use - and not =. - Makes sure the value is xss escaped

vonvick: FIxed

PR #170, opened 2020-11-12
Fix: Add date-fns and date-fns-tz to solve issue of date parsing

vonvick:
#### What does this PR do?
This PR fixes the invalid parsing of ISO-8601 string without timezone

### DIFF
--- a/client/src/js/modules/calendar/views/calendar.js
+++ b/client/src/js/modules/calendar/views/calendar.js
@@ -1,4 +1,4 @@
-define(['marionette', 'backbone', 'dateFnsTz','collections/visits', 'templates/calendar/calendar.html'], function(Marionette, Backbone, DateFnsTz, Visits, template) {
+define(['marionette', 'backbone', 'zonedTimeToUtc','collections/visits', 'templates/calendar/calendar.html'], function(Marionette, Backbone, zonedTimeToUtc, Visits, template) {
     
     // humm
     DISABLE_DAY_SCROLL = false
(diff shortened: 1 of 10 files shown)


client/src/js/modules/calendar/views/calendar.js
```diff
@@ -285,6 +295,7 @@ define(['marionette', 'backbone', 'collections/visits', 'templates/calendar/cale
             this.onScroll = _.debounce(this.onScroll, 100)
             this.collection = new Backbone.Collection()
             this.days = new Backbone.Collection()
+            this.timezone = app.options.get('timezone') ? app.options.get('timezone') : 'Europe/London'
```
drnasmith, line 296: This timezone declaration needs to be in the DayItemView. The child view does not inherit the parent properties.
Also the DayItemView is the only place where the timezone is used in this module, so you can make it a local variable in its initialize method. That way it will have scope within the map functions.

drnasmith (commented):
This is looking good, with a small modification to the timezone declaration (see comments).
Tested setting the php timezone to US and it displays the modified times as expected. Even takes into account british summer time.

JPHall-DLS:
Did you also try changing the time zone of your computer while leaving the PHP timezone set to “Europe/London”?

This would confirm a fix as ECMAScript 6 behaviour is to apply the time zone offset of the user’s computer when evaluating ISO 8601 strings without a timezone. The calendar should reflect the local time at Diamond and not the user’s computer.

vonvick:
@JPHall-DLS I did a test with a different timezone on my computer and I ended up getting the date set to the timezone of the server. So it should work

drnasmith:
At the moment I believe this shows the times local to the users' timezone (taking into account the server timezone correctly). 
For example, when I set the timezone to America/New_York, a session starting at 10:00 (site time) shows in my calendar (Europe tmiezone) as starting at 15:00. That's what I would expect if I wanted to know when a session started - "when do I need to be online to observe the session". However, what we could do with is the ability to show the times as either local to the facility or at the users timezone. I think we merge this fix as is and add the toggle switch as a new feature.

vonvick:
> @JPHall-DLS I did a test with a different timezone on my computer and I ended up getting the date set to the timezone of the server. So it should work

@JPHall-DLS sorry meant to say "displays the time relative to the timezone of the server". @drnasmith is correct.

PR #174, opened 2020-11-20
Fix/add visit

drnasmith:
Allow admin to set visit/session number when creating a new visit.
Initial use case is so we can set a session 0 for new proposals but can be used for any session number.
Checks to make sure there is not session already with the visit number.
Also fixes updating session start/end date.


stufisher (approved):
Looks good, thanks for the fix too

PR #176, opened 2020-11-25
Make prop validator consistent

stufisher:


drnasmith:
Makes sense - deployed onto release branch.

PR #177, opened 2020-11-25
Gridinfo is now linked to dc and not dcg

stufisher:
To be coordinated with database (which has already been updated, but dont know if column has been populated)
This allows multiple grid scans to exist in a group => xray centring for example

I think this should suffice:
```
UPDATE GridInfo
INNER JOIN DataCollectionGroup ON DataCollectionGroup.datacollectiongroupid = GridInfo.datacollectiongroupid
INNER JOIN DataCollection ON DataCollection.datacollectiongroupid = DataCollectionGroup.datacollectiongroupid
SET GridInfo.datacollectionid = DataCollection.datacollectionid
WHERE GridInfo.datacollectionid IS NULL;
```

### DIFF
--- a/api/src/Page/DC.php
+++ b/api/src/Page/DC.php
@@ -1609,7 +1609,7 @@ class DC extends Page
         function _grid_info() {
             $info = $this->db->pq("SELECT dc.datacollectiongroupid, dc.datacollectionid, dc.axisstart, p.posx as x, p.posy as y, p.posz as z, g.dx_mm, g.dy_mm, g.steps_x, g.steps_y, g.pixelspermicronx, g.pixelspermicrony, g.snapshot_offsetxpixel, g.snapshot_offsetypixel, g.orientation, g.snaked
                 FROM gridinfo g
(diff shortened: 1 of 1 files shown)


rjgildea (commented):
@stufisher I think it should be possible to simultaneously support current (`GridInfo.dataCollectionGroupId`) and future (`GridInfo.dataCollectionId`) usage simultaneously, which will avoid making it necessary to coordinate GDA changes across all beamlines, as well as making data analysis/SynchWeb independent of those changes.

rjgildea:
Thanks to @drnasmith we tested my suggested changes on the dev server. After modifying the following `GridInfo` records such that they point to a `dataCollectionId` instead of `dataCollectionGroupId` SynchWeb appears to display them correctly, and even groups them together as a data collection group:

https://cs04r-sc-vserv-163.diamond.ac.uk/dc/visit/cm26458-4/s/20201007/ty/gr
https://cs04r-sc-vserv-163.diamond.ac.uk/dc/visit/cm26458-4/dcg/5545752

Screenshot 2021-03-26 at 10 07 20

Screenshot 2021-03-26 at 10 06 40

PR #179, opened 2020-11-26
return a default value if rotationaxis is not set

stufisher:
With https://github.com/DiamondLightSource/SynchWeb/pull/103 we removed hardcoded omega axis, however some beamlines are not actually populating rotationaxis so need to set a default


drnasmith:
Sensible fix, but some values are empty in the db rather than null! Better than nothing...

PR #182, opened 2020-11-26
sci-9485: dont multi prefix if type already contains multi

stufisher:


drnasmith:
Deployed on release branch

PR #184, opened 2020-11-27
sci-7752: make dials default reprocessing

stufisher:


drnasmith:
Deployed on release branch

PR #185, opened 2020-11-27
i03-428: (2) allow editing of registered container and barcode if still null after dc

stufisher:


drnasmith:
Deployed on release branch

PR #186, opened 2020-11-27
sci-9352: sort containers by name on shipment page

stufisher:
bltimestamp is the same for containers in the shipment, guess there is an update trigger @KarlLevik ? Sort by name on shipment page instead


drnasmith:
Deployed on release branch

PR #187, opened 2020-11-27
i04_1-618: make visit model valid and fix editing comments

stufisher:
I think startdate / enddate are only used for visit management, but i need to check


stufisher:
rebased on master https://github.com/DiamondLightSource/SynchWeb/pull/188

PR #193, opened 2020-12-27
SCI-7442: Update big_ep panel to include all downstream processing

stufisher:
From irakli:

> Updated BigEP panel that reads information from ProcessingJob and AutoProc tables and displays snapshots of the built models.
This change requires replacement of BigEP entries in config.php with
"Big EP" => array('big_ep/', '', '', 'big_ep')

Unions results from autoproc integration and processingjob so will work with both.
Also add unions for autoproc messages so messages can be read for jobs that arent auto proc integration

### DIFF
--- a/api/src/Database/Type/MySQL.php
+++ b/api/src/Database/Type/MySQL.php
@@ -421,6 +421,31 @@ class MySQL extends DatabaseParent implements DatabaseInterface
 
     }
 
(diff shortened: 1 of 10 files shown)


stufisher:
Superseded by https://github.com/DiamondLightSource/SynchWeb/pull/205

PR #194, opened 2021-01-05
feat: Create runs overviews and add Beam Line stats to it

vonvick:
#### What does this PR do?
This PR adds the runs overview menu and also creates a page for displaying a list on beamlines and their stats information


client/src/js/modules/runs/router.js
```diff
@@ -0,0 +1,19 @@
+define(['utils/lazyrouter'], function(LazyRouter) {
+    var Router = LazyRouter.extend({
+        appRoutes: {
+            'runs/overview(/bl/:bl)(/:bl)(/run/:run)(/:id)': 'fetchBeamlinesForRuns'
```
drnasmith, line 4: Think there is an extra (/:bl) in here.

drnasmith: Also I think this should be part of the stats module. So you can add this route and controller to the stats module. You would need to change the view filename to avoid conflicts so "runs_overview.js" view.

While it does have a different route prefix '/runs' instead of '/stats' the page is just a wrapper for beamline stats page. So I think it makes sense for it to live in the stats module.

client/src/js/app.js
```diff
@@ -218,6 +218,7 @@ function(Backbone, Marionette, _, $, HeaderView, SideBarView, DialogRegion, Logi
         'modules/imaging/router',
         'modules/types/em/relion/router',
         'modules/types/em/scipion/router',
+        'modules/runs/router'
```
drnasmith, line 221: See other comments below - I don't think we need a new module for this..

client/src/js/modules/runs/views/overview.js
```diff
@@ -0,0 +1,53 @@
...
+        template: template,
+        className: 'content',
+        regions: {
+            pills: '.beamline-pills',
+            stats: '.beamline-stats'
+        },
+        showBeamlineStats: function(beamline) {
+            var regexExp = new RegExp(/\/bl\/.*/)
+            var url = window.location.pathname
+            if (beamline) {
```
drnasmith, line 22: Please could you provide some comments about why this function needs to use the window push state function.

client/src/js/modules/runs/views/overview.js
```diff
@@ -0,0 +1,53 @@
...
+                name: 'bl',
+                filters: this.beamlines.map(function(beamline) {
+                    return { id: beamline.get('BEAMLINE'), name: beamline.get('BEAMLINE') }
+                })
+            })
+        
+            this.listenTo(this.beamlinePills, 'selected:change', this.showBeamlineStats.bind(this))
+            this.pills.show(this.beamlinePills)
+        },
+        initialize: function() {
```
drnasmith, line 45: It would be nice if this also took into accound the url when it loads.
For example if someone types in /runs/overview/bl/i03 it should load the stats for i03. Similar to how the stats/bl/i03 page works.

drnasmith (commented):
This is good work. I think it will be very useful for some of our scientists.
Detailed comments inline.

PR #198, opened 2021-02-17
Feature: Sample Group View

vonvick:
#### What does this PR do?
This PR creates the ability to select drops in plates for sample view and saving them to the database


client/src/js/app/store/store.js
```diff
@@ -208,6 +207,9 @@ const store = new Vuex.Store({
         })
       })
     },
+    updateLoadingState({ commit }, payload) {
```
drnasmith, line 210: I don't think we need an action to set the loading state when we have a mutation that does the job. It's not an asynchronous task so components can just set the loading flag through the mutation.

api/src/Page/Sample.php
```diff
@@ -1984,24 +1985,45 @@ function _update_sample_group() {
             }
         }
 
-        function _add_sample_to_group() {
-            if (!$this->has_arg('BLSAMPLEID')) $this->_error('No sample specified');
+        function _add_sample_to_group () {
+            $sample = array('BLSAMPLEID' => $this->arg('BLSAMPLEID'), 'BLSAMPLEGROUPID' => $this->arg('BLSAMPLEGROUPID'));
```
drnasmith, line 1989: I think we should have guards at the top. 
So check for presence of BLSAMPLEGROUPID and BLSAMPLEID here with has_arg(), and return error if they are not found.

api/src/Page/Sample.php
```diff
@@ -1984,24 +1985,45 @@ function _update_sample_group() {
...
+
+                array_push($collection, $sample_update_result);
+            }
+
+            $this->db->end_transaction();
+            $this->_output($collection);
+        }
+
+        function _save_sample_to_group($sample_data) {
+            if (!isset($sample_data['BLSAMPLEID'])) $this->_error('No sample specified');
```
drnasmith, line 2010: Rather than an error here it would be better to return null.
Then all the return paths (_error  or _output) are in the calling function which is cleaner.
You would need to check the return value in bulk_update to error if one of the calls to _save_sample_to_group returns null and then _error()

api/src/Page/Sample.php
```diff
@@ -2022,13 +2044,13 @@ function _add_sample_to_group() {
 
             if ($chk[0]['SAMPLES'] != $in_proposal) $this->_error('No such sample group');
 
-            $order = $this->has_arg('GROUPORDER') ? $this->arg('GROUPORDER') : 1;
-            $type = $this->has_arg('TYPE') ? $this->arg('TYPE') : null;
+            $order = isset($sample_data['GROUPORDER']) ? $sample_data['GROUPORDER'] : 1;
```
drnasmith, line 2047: It looks like GROUPORDER and TYPE are not set in line 1989 above?

api/src/Page/Sample.php
```diff
@@ -2022,13 +2044,13 @@ function _add_sample_to_group() {
...
+            $type = isset($sample_data['TYPE']) ? $sample_data['TYPE'] : null;
 
             $this->db->pq("INSERT INTO blsamplegroup_has_blsample (blsampleid, blsamplegroupid, grouporder, type) 
                 VALUES (:1,:2, :3, :4)", array($this->arg('BLSAMPLEID'), $sgid, $order, $type));
 
-            $this->_output(array('BLSAMPLEGROUPSAMPLEID' => $sgid.'-'.$this->arg('BLSAMPLEID'), 'BLSAMPLEID' => $this->arg('BLSAMPLEID'), 'BLSAMPLEGROUPID' => $sgid));
+            return array('BLSAMPLEID' => $this->arg('BLSAMPLEID'), 'BLSAMPLEGROUPID' => $sgid);
```
drnasmith, line 2053: Need to check if this still works with the current branch. We are not returning the hybrid samplegroupsampleid?

Also still using this->arg() here not sample_data. Not convinced the use of sample_data saves much here.  You still need to check presence of keys in the array which is what this->has_arg() does. 

Using the args/has_arg is consistent with the rest of the code.

An alternative is to setup the parameters in the calling function and pass them in as variables. So create a $blsampleId, $blsampleGroupId, $groupOrder, $type as arguments to _save_sample_to_group(). Then you define a clear method signature.

Checkout the _update_autocollect_session method in Proposal.php for an example

drnasmith (commented):
Comments inline...

api/src/Page/Sample.php
vonvick: fixed

api/src/Page/Sample.php
vonvick: @drnasmith since we are also using this for the bulk update which is run as a DB transaction, do we still need to return null or throw an error to rollback the transaction?

drnasmith: Yes, return null from _save_sample_to_group. That method will return data or null.
Then in the calling function _bulk_update_samples_in_group, check for the null response.
If null throw the error.
_bulk_update_samples_in_group will either call _output or _error.

I note that _do_add_sample uses the same pattern as you had - i.e. throws errors in the middle of the code. I think its cleaner to use functions that return data or null, then check their result. It keeps all the return paths at the same level. Does that make sense?

PR #204, opened 2021-03-11
Declare provenance when sending messages into Zocalo

Anthchirp:
Currently when messages are submitted via `dlstbx.go` [we identify the sending user and host in the message headers](https://github.com/DiamondLightSource/python-dlstbx/blob/393eab8b7f105150533858d443bc9c87d667f7e9/src/dlstbx/cli/go.py#L148-L149) to aid debugging:
```python
        {
            "dlstbx.go.user": getpass.getuser(),
            "dlstbx.go.host": socket.gethostname(),
        }
```

It would be great if Synchweb could do the same. I obviously have no clue what interesting and relevant information you could pass on. I suppose the logged on user and the hostname as `synchweb.user` and `synchweb.host` would be most useful.

Unfortunately I have no idea about the Synchweb internal structure and where to get that information, and my PHP skills are way out of date. Which is why in this PR I only make an absolute minimal change.

### DIFF
--- a/api/src/Queue.php
+++ b/api/src/Queue.php
@@ -27,7 +27,8 @@ class Queue
                 $queue,
                 json_encode($message, JSON_UNESCAPED_SLASHES),
                 array(
-                    'persistent' => ($persistent === true)
+                    'persistent' => ($persistent === true),
+                    'source' => 'synchweb'
                 )
             );
 


Anthchirp:
(just to clarify: `synchweb.host` would be the Synchweb _server_ hostname, not the client's)

drnasmith:
It is easy enough to get hold of the user info (i.e. login/fedid).
As for "source/host" I think making that a config property is better.
In future the php service could be running behind an API gateway so hostname or IP address would not be meaningful.
There is already a config section for zocalo connections so we can just add to that.

Anthchirp:
The important bit for us is that the message comes from synchweb. The particular hostname/IP would be more for you in case something needs to be cross-correlated with logs. Soo - entirely your choice what to put there

andy-preston:
I've got a possible implementation of this here:
https://github.com/andy-preston/SynchWeb/commit/8ee32dbf3903a5b22eb09b11cabebc19f87935a9

Anthchirp:
Synchweb is now reporting the fact that the message comes from Synchweb, without any further information.

andy-preston:
Looking at these:
https://github.com/DiamondLightSource/SynchWeb/blob/master/api/src/Queue.php 
https://github.com/DiamondLightSource/SynchWeb/blob/c2b8e9554ff983c16d2fec338e9d815787f7b0c5/api/src/Page/EM.php#L872
which are live in the current release of SynchWeb ... you should be seeing host name and user ID.
It you don't this is still a bug and needs fixing.

PR #205, opened 2021-03-22
Auto/downstream processing plugin refactoring

stufisher:
Removes all status file scraping, status all come from db now. For autointegration that requires a link to autoprocintegration, for downstream an autoprocprogram must have a processingjob entry. 

This can be used in any discipline now to show automated processing. Any processing types that dont have a relevant plugin will be shown as a table with start / stop and a list to associated files / archive download

Remove all specialisation for viewing maps / models for different jobs, files are all retrieved via autoprocprogramattachments.
Remove all specialisation for downloading job archives, use autoprocprogramattachment.

Added processing status for saxs and xpdf.

This will need some extensive testing on staging as i cant test everything on dev.

https://ispyb.diamond.ac.uk/dc/visit/nt28218-3/id/6140309

https://ispyb.diamond.ac.uk/dc/visit/nt28218-3/id/6140318
Screenshot 2021-04-10 at 13 21 34


client/src/js/modules/types/saxs/dc/dc.js
```diff
@@ -1,24 +1,33 @@
...
         },
         
         showDiff: function(e) {
             e.preventDefault()
             this.$el.find('.diffraction a').eq(0).trigger('click')
         },
+
+        loadAP: function(e) {
+            if (!this.ap) {
+              this.ap = new DCDownstreamView({ id: this.model.get('ID'), el: this.$el.find('div.downstream') })
```
drnasmith, line 28: This feels odd now. It's autoprocessing but using the downstream processing api. Presume this is because saxs uses processingjob as the link table rather than autoprocintegration. However, if saxs (or other types) extend to register jobs as both automated and downstream this gets confusing. Will this work scale to support detecting which jobs are downstream and which are reprocessed?

drnasmith (changes requested):
So lots of code to look through. 
Its running on ispyb-staging.
After some tweaks to integrate with latest master candidate it works. However, BigEP jobs for XDS and DIALS do not seem to be reflected correctly. Using the link you provided, compare staging and production ispyb.
Also we need to get xpdf integrated before deploying.
The autoproc and dowstream status bar will need some UI tweaks. Maybe sum the number of results rather than show a large list of icons - plus a separator between processing applications.

api/src/Downstream/Type/autosharp.php
```diff
@@ -0,0 +1,9 @@
+<?php
```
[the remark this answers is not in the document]
stufisher: They can be title but not pascal as they are auto loaded from the name in the database

api/src/Downstream/Type/fast_ep.php
```diff
@@ -0,0 +1,127 @@
...
+        $dat = array();
+
+        $ats = array();
+        $pdb = $this->_get_attachments("sad_fa.pdb");
+        if ($pdb) {
+            if (file_exists($pdb["FILE"])) {
+                $pdb = file_get_contents($pdb["FILE"]);
+                foreach (explode("\n", $pdb) as $l) {
+                    if (strpos($l, 'HETATM') !== false) {
+                        $parts = preg_split('/\s+/', $l);
```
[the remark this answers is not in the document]
stufisher: These are the heavy atom positions that have been found

client/src/js/modules/types/saxs/dc/dc.js
stufisher: Indeed this is just the way the database works. In ISPyB autoprocessing = integration and this is very mx specific. Any other technique will have to use "downstream". Its better that than renaming 15 tables. Theres no concept of "downstream" for other disciplines right now

stufisher: Reprocessed yes, but they will be reprocessed "downstream" for other techniques. The status ticks only show jobs with automatic = 1. But reprocessed will show in the tab, and we add a marker to show they are not automatic

stufisher:
I've already spotted a few bugs reviewing the pr. I'll also rebase on master

client/src/js/modules/types/saxs/dc/dc.js
stufisher: And for users they will obviously be marked as showsn as autoprocessing even if they use the "downstream" code

stufisher:
> BigEP jobs for XDS and DIALS do not seem to be reflected correctly. Using the link you provided, compare staging and production ispyb.

This pr also splits up big ep into their individual jobs (they were merged into a single tab on prod), which is why they look different. You'll now get a tab for each of the procesed (crank, autobuild, autosharp). The images will be back when @isikhar gets a pr merged into dlstbx to put the images on each job individually and the top settings table will be persisted into processingparameters.

stufisher:
Mostly addressed your points. Awaiting another attachment for big ep and i will put the plots back

stufisher:
Jobs now indicate if they were manually triggered with a refresh icon, works for both auto and downstream proceses

stufisher:
to check:
 * sanitise job name from archive download
 * add *predict upstream integrator (needs a plugin)

stufisher:
Group by program type to eliminate having 12+ tabs

PR #207, opened 2021-03-25
Filter out sub samples with source!=manual for now

KarlLevik:
We are now (in theory at least) storing sub-samples that have been generated by CHiMP, but we currently don't have any UI for these. 

Therefore, for now let's just modify the relevant queries so they only select sub samples submitted by humans, i.e. `blsubsample.source='manual'`. 

Note that 'manual' is the default value for the column, and all the historical rows in the table have this value.


drnasmith:
On release branch at DLS

PR #213, opened 2021-04-05
dont allow marking shipment sent if a dewar is processing

### LINKED ISSUE #201: Shipment "Mark as Sent" button can break scanned Assign Containers view
If a user clicks the "Mark as Sent" button in the Shipment view for a shipment when a container from a dewar from that shipment has been assigned to a sample changer location, it will cause the status of the dewar to be changed from `processing` to `sent to facility`!  This is disastrous because it will make the container disappear from the scanned Assign Containers view because the dewar status is no longer `processing` which is one of the conditions for querying which containers are in the sample changer!  Any other database queries that rely on the dewar status being `processing` as part of finding which containers are in the sample changer will also break.  It should not be possible for a user to break things like this simply by clicking the "Mark as Sent" button of a shipment.

shipment-mark-as-sent

stufisher:
Closes #201

### DIFF
--- a/api/src/Page/Shipment.php
+++ b/api/src/Page/Shipment.php
@@ -263,7 +263,8 @@ class Shipment extends Page
                 if (array_key_exists($this->arg('sort_by'), $cols)) $order = $cols[$this->arg('sort_by')].' '.$dir;
             }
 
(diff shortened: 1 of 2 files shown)


John-Holt-Tessella:
Did we need to work out how the shipping status got set to pickup booked, or the awb created without leaving processed?

John-Holt-Tessella (approved):

PR #215, opened 2021-04-06
Changes as suggested in SCI-9600 in comment of 19. Feb. 2021

KarlLevik:
These changes should make the "inspections" query on the imaging page execute much faster for the default parameters.

You will probably have to make some changes to the front-end to handle the removal of the "imagesscored" field.

These are the changes in more detail:

- I've moved both JOINs with InspectionType and ScheduleComponent into sub-queries.
- I've removed the "GROUP BY i.containerinspectionid" and the "LEFT OUTER JOIN BLSampleImage ...".
- I've put the min function into a subquery: (SELECT DATE_FORMAT(min(modifiedtimestamp), '%d-%m-%Y %H:%i') FROM BLSampleImage WHERE containerinspectionid = i.containerinspectionid AND blsampleimagescoreid IS NOT NULL)
- and removed the "imagesscored" field / count function as I don't think it's strictly needed since it's 1 when the min function is a value and 0 when it's NULL. I.e. you can derive it from the "imagesscoredtimestamp" field.

This gives me an execution plan with "Using filesort", but no "Using temporary", and it executes in less than a second on our production database cluster. (I can't guarantee it will be as fast elsewhere.)

### DIFF
--- a/api/src/Page/Imaging.php
+++ b/api/src/Page/Imaging.php
@@ -388,17 +388,16 @@ class Imaging extends Page
                 if ($this->arg('ty') == 'MANUAL') $where .= " AND i.manual=1";
             }
 
+            $groupby = "";
             if ($this->has_arg('delta')) {
+                $groupby = "GROUP BY i.containerinspectionid";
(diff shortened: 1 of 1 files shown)


drnasmith:
Tested this but it removes some fields that VMXi need, so won't work in this form.

PR #221, opened 2021-04-27
php7 will return warning for sizeof on none array

stufisher:


drnasmith:
On release branch at DLS

PR #228, opened 2021-05-14
Feature: Sample Group View

vonvick:


vonvick:
@morybc this is the sample group management work I am currently doing for VMXi and MX (partially). I have been keeping it up to date with the master branch so it should not hard to integrate with the current version of synchweb.

PR #239, opened 2021-08-17
Integrate/imca shipments

drnasmith:
This merges the imca shipments branch with vue changes.
It will need to coexist with the mx-add-container page that Victor is working on. 
Potentially there should be a switch in the ContainerAddWrapper component that can select which page is loaded...


drnasmith:
Also excuse the reverts - the uas changes were supposed to be on a separate branch...

stufisher:
This ones mine, and in use at IMCA-CAT. It can be left alone, not ready to be merged into master any time soon (please dont delete it!)

PR #245, opened 2021-09-14
visit regexp is \w+ not \w\w for proposalcode, db is varchar(45)

stufisher:
Other facilities use proposalcode > 2 chars and db allows this


stufisher:
Will need to test this on staging to check it doesnt break anything

stufisher:
It would be better to do something like this tbh: https://github.com/DiamondLightSource/SynchWeb/blob/master/client/src/js/utils/validation.js

where the regexps can be reused

ndg63276:
Closing PR as it is very old. Branch will be deleted in 1 month unless someone objects.

PR #247, opened 2021-09-16
fix sample editing

stufisher:
Closes https://jira.diamond.ac.uk/browse/SCI-10049


NigelWilson (approved):

John-Holt-Tessella:
Also closes LIMS-78

api/src/Page/Sample.php
```diff
@@ -1928,8 +1925,11 @@ function _update_sample()
             $dcps = $this->db->pq("SELECT dataCollectionPlanId FROM BLSample_has_DataCollectionPlan
                                 WHERE blSampleId = :1", array($this->arg('sid')));
 
-            if (sizeof($dcps)) {
-                foreach ($dcps as $dcp) {
+            if(sizeof($dcps)) {
+                $maxLocation = $this->_get_current_max_dcp_plan_order($this->args['CONTAINERID']);
```
ndg63276, line 1929:
```suggestion
```
Is this line needed?

api/src/Page/Sample.php
```diff
@@ -1928,8 +1925,11 @@ function _update_sample()
...
+                $maxLocation = array_key_exists('CONTAINERID', $this->args) ? $this->_get_current_max_dcp_plan_order($this->args['CONTAINERID']) : array(0);
+
```
ndg63276, line 1931:
```suggestion
                $maxLocation = sizeof($maxLocation) ? $maxLocation : -1;
```
Don't we need this line?

api/src/Page/Sample.php
John-Holt-Tessella: No but the line `$maxLocation = $maxLocation ? $maxLocation + 1 : 0;` but I didn't like it. So instead I have moved all the code for determining if there is alread an plan order into that function and called it get next index.
Could you review and I will put it on ipsyb-dev-5

ndg63276 (approved):
Works for me

PR #248, opened 2021-09-16
Update LDAP.php to support start_tls

dsclassen:
Add `ldap_start_tls` because my LDAP server requires a secure connection. This may break LDAP authentication with other LDAP configurations so perhaps it would be better to add start_tls = true/false to the config.php file and then do some sort of if/else switch in LDAP.php?


morybc:
Recreated the PR, with additional config parameter, to switch on/off behaviour - see https://github.com/DiamondLightSource/SynchWeb/pull/332**

Thanks for your contribution Scott.

PR #250, opened 2021-10-04
[SCI-10071] marionette-wrapper allows multiple Marionette views per Vue scope

andy-preston:


vonvick:
My bad 😞 . I think the PR looks good to be merged then.

PR #257, opened 2021-11-03
Enable triggering Zocalo/AlphaFold on adding new protein sequence

rjgildea:
If a new protein is added with a sequence, or an existing protein is updated to add a sequence (where the sequence was previously NULL/empty), then submit Zocalo recipes defined by the `zocalo_recipes_on_add_protein_sequence` config variable.

Add a generic `_submit_zocalo_recipe()` function to the base `Page` class.

### DIFF
--- a/api/config_sample.php
+++ b/api/config_sample.php
@@ -105,9 +105,9 @@
 
     # This is used to trigger Zocalo recipes on adding new Protein sequences
     # Set to empty string to disable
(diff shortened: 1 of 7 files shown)


morybc:
@JPHall-DLS could you review this please?  It looks reasonably stand-alone so hopefully shouldn't impact too much?

ndg63276:
No longer needed according to Graeme.

PR #302, opened 2022-06-23
Feature: Prototype H5Web and HSDS as a service on K8s

vonvick:
####  What does this PR do?
This PR builds a link to launch a MAXIV like viewer with a H5Web client and HSDS server that can be launched from SynchWeb and run in K8s


John-Holt-Tessella:
I am not sure what this PR is but it was pointing at an old release branch so have pointed it at prerelease

ndg63276:
Closing as too old to investigate.

PR #313, opened 2022-07-14
Feature/add podman release support

morybc:
Added new support for prod deployment via podman container.


John-Holt-Tessella (approved):
Have run container using the dev php file on my local system. It starts and I can see the login page. I can not yet log into the system to test any more funcationality)

PR #320, opened 2022-08-05
Changed path to conexs backend and added one reference paper

Villtord:
Changed path to conexs backend and added one reference paper


Villtord:
Just for future reference (if any) I will also reply here. That commit was about to improve web-conexs project frontend, in general I would say it is still relevant. But as I don't work in Data Analysis anymore I am not maintaining this project. Please ask J.Filik about it.

ndg63276:
Closing as planning a full rewrite when HTCondor is turned off.

PR #322, opened 2022-08-14
fix regex expressions for PCRE2 (i.e. for PHP >= 7.3)

dsclassen:
If you want to host SynchWeb on a web server running PHP >= 7.3 where the underlying regex libraries have changed from PCRE to PCRE2 then you will need the changes in this PR.

You can [read](https://php.watch/versions/7.3/pcre2) about the differences between PCRE and PCRE2 and what affect it may have on regex code.

This [utility](https://regex101.com) is a good one for helping evaluate regex.

I have fixed a few of the problematic expressions while hopefully retaining the intended behavior.

### DIFF
--- a/api/src/Page.php
+++ b/api/src/Page.php
@@ -35,7 +35,7 @@ class Page
                                       'sort_by' => '\w+',
                                       'order' => 'desc|asc',
                                       'ty' => '\w+',
-                                      's' => '[\w\s-]+',
+                                      's' => '[\w\s\-]+',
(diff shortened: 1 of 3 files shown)


dsclassen:
I believe these changes should be backwards compatible with earlier versions of PHP.

dsclassen:
Maybe this PR is made obsolete by #329

dsclassen:
It seems efforts are underway to make SynchWeb compatible with PHP>=7.4 so I'll close this PR.

PR #332, opened 2022-09-12
Extend dsclassen PR248 - add in config property to enable LDAP over TLS

morybc:
Extend https://github.com/DiamondLightSource/SynchWeb/pull/248 to add config option to switch on/off TLS use for LDAP (default to off, to keep current behaviour as standard).

Apologies, Scott, for recreating your work - wasn't sure how to add to your forked PR, so thought it was just easier this way.


John-Holt-Tessella (approved):
Code looks good but I can not test because I do not have access to LDAP

PR #342, opened 2022-09-23
Feature/add js unit tests

morybc:
Add test framework, vue-testutils, to test the Vue code + update CI to include and add a couple of initial test classes to demonstrate use of mocks, etc.

Note, have not added more general tests for the older javascript code as am assuming this will gradually be migrated into vue - so the related logic can be tested at that point.


John-Holt-Tessella (approved):
This looks like it works, a few minor changes added

PR #345, opened 2022-10-03
Feature/update dhl api

morybc:
Upgrade DHL-API to use latest version + add some simple php unit tests.

Note, this includes all the previous work for the 7.4 upgrade - which has yet to be merged.  Only the last three commits directly relate to this change.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ on:
 defaults:
   run:
     shell: bash
-    working-directory: ./api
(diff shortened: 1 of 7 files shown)


John-Holt-Tessella:
Merge in previous commits into pre-release branch now there are outstanding commits.

I want to merge master in first and then we can start using this.

ndg63276:
Closing PR as we now use the shipping service and new DHL API. Branch will be deleted in one month unless someone objects.

PR #346, opened 2022-10-03
Add support for Microsoft Active Directory in LDAP authentication

stuartcampbell:
The current schema / format of the `dn` that works with OpenLDAP servers does not always work when trying to use an Active Directory domain controller as an LDAP authentication server.  

In testing it seems more reliable to authentication using a username of `DOMAIN\\username` rather than trying to guess what the `dn` is being used at a given site.  The problem is that the username is not always part of the `dn` for AD, but using the username and domain name always seems to work.  

I've tested the code in this PR at BNL using the onsite Active Directory Domain Controllers (and also using a DUO MFA proxy server).  I have not tested against Azure AD.


api/src/Authentication/Type/LDAP.php
```diff
@@ -39,14 +42,27 @@ function authenticate($login, $password)
             if ($ldap_use_tls) {
                 ldap_start_tls($conn);
             }
-
+
```
ndg63276, line 45:
```suggestion

```

api/src/Authentication/Type/LDAP.php
```diff
@@ -39,14 +42,27 @@ function authenticate($login, $password)
...
-                // DN syntax: "uid=<login>,ou=people,dc=example,dc=com"
-                return ldap_bind($conn, "uid=" . $login . "," . $ldap_search, $password);
+                if ($ldap_server_type == "activedirectory") {
+                    if (!$active_directory_domain) {
+                        error_log("'active_directory_domain' parameter is not defined.");
+                        error_log("\t This is required when LDAP server type is 'activedirectory'");
+                        return false;
+                    }
+                    $ldap_user = $active_directory_domain . "\\" . $login;
+		        } else {
```
ndg63276, line 54:
```suggestion
                } else {
```

NKatti2011:
Looks good to me

PR #347, opened 2022-10-12
Fix: SSX Auto Processing

vonvick:


John-Holt-Tessella:
Already added somewhere else and is in aster

PR #350, opened 2022-10-21
Feature/adjust code structure

morybc:


api/index.php
```diff
@@ -29,56 +30,107 @@
...
+$app->container['dispatch']->dispatch();
 
-// MySQL database class hard-coded
-$db = new MySQL($isb['user'], $isb['pass'], $isb['db'], $port);
-
-// Alternatively, use dynamic class instantiation.
-// Database type ($dbtype) specified in config.php.
-// $db = Database::get();
-
-$db->set_app($app);
```
John-Holt-Tessella, line 68: This is missing from the new code and is needed because a content header is set depending on it being set. We should add the app to the constructor.

John-Holt-Tessella: It isn't, and it is hard to add it to the constructor so I have moved it into the DI code.

api/src/Controllers/UserController.php
```diff
@@ -0,0 +1,355 @@
...
+        }
+        else
+        {
+            $this->_output(array('USERGROUPID' => $this->userData->addGroup($this->arg('NAME'))));
+        }
+    }
+
+    function _update_group()
+    {
+        $this->haltIfLackingPermission('manage_groups');
```
John-Holt-Tessella, line 149: Missing check that group exists. We might not care think more deeply about this

api/src/Controllers/UserController.php
```diff
@@ -0,0 +1,355 @@
...
+        $this->userData->updateGroup($this->arg('gid'), $this->arg('NAME'));
+        $this->_output(array('NAME' => $this->arg('NAME')));
+    }
+
+    function _add_group_permission()
+    {
+        $this->haltIfLackingPermission('manage_groups');
+        $this->userData->addGroupPermission($this->arg('gid'), $this->arg('pid'));
```
John-Holt-Tessella, line 157: missing check that group and permission exist

api/src/Controllers/UserController.php
```diff
@@ -0,0 +1,355 @@
...
+                'data' => $rows,
+            ));
+        }
+    }
+
+
+    function _add_permission()
+    {
+        $this->haltIfLackingPermission('manage_perms');
+        $this->_output(array('PERMISSIONID' => $this->userData->addPermission($this->arg('TYPE'), $this->argOrEmptyString('DESCRIPTION'))));
```
John-Holt-Tessella, line 341: Missing checks that gid and pid exists

PR #351, opened 2022-11-02
Feature/continue refactoring

morybc:
First part of refactoring to introduce a controller and data layer + associated unit tests.  Focuses on Admin code which is a hidden piece of functionality not available to general users.


api/index.php
```diff
@@ -29,56 +32,117 @@
...
-// MySQL database class hard-coded
-$db = new MySQL($isb['user'], $isb['pass'], $isb['db'], $port);
-
-// Alternatively, use dynamic class instantiation.
-// Database type ($dbtype) specified in config.php.
-// $db = Database::get();
-
-$db->set_app($app);
-
-$auth = new Authentication($app, $db);
```
John-Holt-Tessella, line 70: Got here with checking for after Christmas

api/src/Controllers/AssignController.php
```diff
@@ -0,0 +1,149 @@
...
+            $this->assignData->assignContainer($cs[0], $this->arg('pos'));
+            $this->_output(1);
+        }
+        else
+        {
+            $this->_output(0);
+        }
+    }
+
+    function unassignContainer()
```
John-Holt-Tessella, line 47: Test history displays if unassign container because it now has a blank string in

John-Holt-Tessella (approved):

api/src/Controllers/AssignController.php
John-Holt-Tessella: Tested,, display is good

PR #352, opened 2022-11-02
Feature/refactor calendar code

morybc:
Continue refactoring - extracting out controller and data layer classes from Cal.php page.  Note, these have not been tested in the live application as I could not figure out where they were actually exposed!


ndg63276:
Closing PR as it is very old. Branch will be deleted in one month unless someone objects.

PR #353, opened 2022-11-02
Feature/refactor contact code

morybc:
WIP - refactoring contact code.  This is nearly complete although the unit tests have not been written yet.  Also, the structure needs changing to make use of the UserController class to avoid code duplication.  This can be injected into the main class and reused directly, I think...


ndg63276:
Closing PR as it is very old. Branch will be deleted in one month unless someone objects.

PR #354, opened 2022-11-16
[LIMS-130]Fix: optional local contact for dispatch

MattPrit:
**JIRA ticket:** [LIMS-130](https://jira.diamond.ac.uk/browse/LIMS-130)

**Changes:**
- Make local contact and Dewar location optional fields of the Dewar dispatch form.
- Make the local contact field read-only

**To test:**
- Check that the form can be submitted both with and without the local contact and dewar location fields populated
- Check that the dispatch email(s) are sent correctly in all cases


JPHall-DLS (approved):
LGTM

PR #358, opened 2022-11-16
[LIMS-218]Fix: Add support for tracking by DHL piece number

MattPrit:
**JIRA ticket:** [LIMS-218](https://jira.diamond.ac.uk/browse/LIMS-218)

**Changes:**

- Add support for DHL tracking requests using the piece number (LPNumber) rather than the airway bill number (AWBNumber)
- Small refactor of tracking backend
- Raise errors if attempting to track a non-DHL shipment or tracking number doesn't match any of the expected patterns

**To test:**

- For a given DHL shipment, check that tracking history can be received for both piece numbers and airway bill numbers being used as the tracking numbers to/from the facility.

### DIFF
--- a/api/src/Page/Shipment.php
+++ b/api/src/Page/Shipment.php
@@ -894,61 +894,82 @@ class Shipment extends Page
                 $args = array($this->arg('DEWARID'));
             }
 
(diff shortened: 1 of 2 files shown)


JPHall-DLS (approved):
LGTM

PR #359, opened 2022-11-16
[LIMS-200]Fix: Fix shipping label address

MattPrit:
**JIRA ticket:** [LIMS-200](https://jira.diamond.ac.uk/projects/LIMS/issues/LIMS-200)

**Changes:**

- Trim trailing whitespace from laboratory address before adding it to the shipping label
- Change shipping label address line breaks from being appended to prepended to address lines

**To test:**

- Check that addresses on shipping labels are formatted correctly for addresses of all allowed lengths


JPHall-DLS (approved):
LGTM

PR #361, opened 2022-11-17
[LIMS-93]Fix: Change logic of dewar dispatch courier fields

MattPrit:
**JIRA ticket:** [LIMS-93](https://jira.diamond.ac.uk/browse/LIMS-93)

**Changes:**

- Change logic of Dewar dispatch form such that: If the facility courier account is used, courier details fields are hidden. If the facility account is not used and courier details have been fully specified, then courier details fields are shown disabled by default, with the option for editing. If the facility account is not used but courier details are not fully specified, then the fields are enabled by default, with no option to disable.

**To test:**

- Check that, for each combination of provided courier account information, the correct fields are visible.

### DIFF
--- a/client/src/js/modules/shipment/views/dispatch.js
+++ b/client/src/js/modules/shipment/views/dispatch.js
@@ -91,11 +91,13 @@ define(['marionette', 'views/form',
                 self.updateLC()
             })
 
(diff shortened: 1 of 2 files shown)


JPHall-DLS (approved):
LGTM

PR #366, opened 2022-11-22
[LIMS-231]Improvement: Update and move description on shipping label

MattPrit:
**JIRA ticket:**

**Changes:**

- Swap positions of 'This way up' and shipment description.
- Update the description to read 'Cryogenic dry shipper. Not restricted, ADR A346, IATA A152.'
- Change validation on package description field on the create air waybill page to allow the new description.

**To test:**

- Check that shipment labels are formatted as expected.
- Check that airway bills can be created using the new description.

**Note:** To update the description on airway bills, `$package_description` will need to be updated in `config.php`.


JPHall-DLS (approved):
LVGTM

PR #367, opened 2022-11-23
R2.4.1 (Fix): Remove NumberOfPieces from DHL shipment request

MattPrit:
This fixes an issue made when merging DHL changes into the R4.2.1 branch: NumberOfPieces was not removed from DHL shipment requests - this is required by the new version of the DHL XML API schema.


JPHall-DLS:
Fixed elsewhere, no longer required.

PR #388, opened 2022-12-15
LIMS-225: Allow hyphens in user path when creating pucks

ndg63276:
**JIRA ticket: [LIMS-225](https://jira.diamond.ac.uk/browse/LIMS-225)**

**Changes:**

- Change regex validation on User Path field to allow hyphens before and after the forward slash

**To test:**

- Check the form can be submitted with hyphens in the User Path
- Check the sample can be edited to add hyphens to the User Path
- Check the Bulk Editor can put hyphens in the User Path


MattPrit:
Closed accidentally through branch deletion

PR #391, opened 2022-12-21
Pass in class context to `._each` loop and fix function call

sshussain270:


ndg63276:
Ticket at https://jira.diamond.ac.uk/browse/LIMS-143

PR #435, opened 2023-02-02
[LIMS-212] Fix: Persist edited visit comments

NKatti2011:
JIRA ticket: [LIMS-212](https://jira.diamond.ac.uk/browse/LIMS-212)

Changes:

Fix such that the visitor page persists comments when editing the comment field. Have removed visits page from marionette framework and replaced this with vue. 

To test:

- Check user journey is the same.

- Check ISPYB is capturing change in visitor comments.

### DIFF
--- a/client/src/js/modules/visits/views/visit_list.vue
+++ b/client/src/js/modules/visits/views/visit_list.vue
@@ -39,6 +39,7 @@
                                 <p v-if="value.key != 'COMMENTS' && value.key != 'ARCHIVED'">{{visit[value.key]}}</p>
                                 <p v-if="value.key == 'COMMENTS' && visit.clicked == false">{{visit.COMMENTS}}</p>
                                 <input v-if="value.key == 'COMMENTS' && visit.clicked == true"
+                                       title="Comment cannot be seen by User Office"
                                        v-model="visit.edited_comment"
                                        v-on:keyup.enter="onEnter(visit)" />
                                 <i class="fa fa-archive r" title="This visit is archived, file is no longer available on disk" v-if="value.key == 'ARCHIVED' && visit.ARCHIVED == 1"></i>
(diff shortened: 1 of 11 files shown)


client/src/js/modules/visits/views/visit_list.vue
```diff
@@ -0,0 +1,216 @@
...
+                    <tr class="hover:tw-bg-table-row-hover-background"
+                        @click="navigator($event, visit)"
+                        @focusout="handleFocusOut(visit)"
+                    :class="{
+                                'tw-bg-table-body-background-odd' : rowIndex % 2,
+                                'tw-bg-content-sub-header-background' : isOdd % 2,
+                            }"> 
+                        <td v-for="(value) in headers" :key="value.id"
+                            :id="value.key" class="tw-p-1 tw-text-center">
+                                <p v-if="value.key != 'COMMENTS' && value.key != 'ARCHIVED'">{{visit[value.key]}}</p>
```
gfrn, line 39: Always use [strict equality (===)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Strict_equality), even when it doesn't seem like much of a difference, since Javascript does funny things with type conversions and we could be introducing bugs with non-strict equality checks

client/src/js/modules/visits/views/visit_list.vue
```diff
@@ -0,0 +1,216 @@
...
+
+        },
+        async onEnter(visit) {
+            // when changing comments for visits
+            visit.COMMENTS = visit.edited_comment;
+
+            visit.clicked = false;
+
+            await this.$store.dispatch('updateDataToApi', {
+            url: '/proposal/visits/' + visit.VISIT + '?prop=' + visit.PROPOSAL + '&COMMENTS=' + visit.COMMENTS,
```
gfrn, line 183: Maybe consider using [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) instead of string concatenation, as it would result in easier to read code. For example, this would become:

`/proposal/visits/${visit.VISIT}?prop=${visit.PROPOSAL}&COMMENTS=${visit.COMMENTS}`

PR #446, opened 2023-02-07
[LIMS-377] SynchWeb: Display resolution where I/SigI == 2 in frontend

NKatti2011:
Ticket: [LIMS-377](https://jira.diamond.ac.uk/browse/LIMS-377)

Display resIOverSigI2 value in SynchWeb in dataset stats view, where it only shows when it is available in ISPYB.


MattPrit (approved):
Looks good to me

PR #458, opened 2023-02-10
[LIMS-321]Fix: 

MattPrit:
JIRA ticket: [LIMS-321](https://jira.diamond.ac.uk/browse/LIMS-321)

Changes:

- Add the call to `countries.fetch` to the array awaited on rendering the dispatch view.

To test:

- Check that the country drop-down is now always populated with the complete list of countries, rather than just 'United Kingdom/other'

Note: It is not really possible to ensure that the list is *always* populated. When testing this change I added an artificial delay in the back end function `_get_countries`. Before this change the delay would result in the list of countries not being populated; after the change the list would be populated once the request returned.


NKatti2011 (approved):
LGTM

PR #472, opened 2023-02-14
LIMS-449: Left-align column headings in data processing tables

ndg63276:
Ticket: [LIMS-449](https://jira.diamond.ac.uk/browse/LIMS-449)

* Left align table headings (<th> tags) for data processing tables, otherwise the other rows look offset
* Change I/sig(i) to I/sig(I)


NKatti2011:
LGTM.

PR #475, opened 2023-02-15
LIMS-453: Email container owners when a dewar returns to the dewar store

ndg63276:
Ticket: [LIMS-453](https://jira.diamond.ac.uk/browse/LIMS-453)

Currently when dewars are returned to the Randex we send an email to the lab contact to say the dewar is ready for return or internal transfer. Feedback from users is that the current notifications aren't going to the right people, and that the people creating the samples don't get told once their samples are done. This PR adds all container "owners" to the CC list of the email.


NKatti2011:
LGTM.

PR #476, opened 2023-02-15
LIMS-342: Move MCA peak labels when window resizes

ndg63276:
Ticket: [LIMS-342](https://jira.diamond.ac.uk/browse/LIMS-342)

If the browser window is resized, the peak labels (eg "Se-K") on an MCA spectrum should move with the peaks themselves.
Also fix typo on "opacity".


ndg63276:
It was asked if the `if (o.left > 0)` clause is necessary, but without it, peaks off to the left of the graph are labelled, eg
image

NKatti2011:
LGTM.

PR #477, opened 2023-02-16
Lims 421/bug/insert lab info

John-Holt-Tessella:
If a lab record does not exist and then a value is inserted the previous code failed. This now works because it only updates the values which it has to set


api/src/Database/DatabaseQueryBuilder.php
```diff
@@ -71,30 +73,62 @@ public function whereIdEquals($columnName, $value) {
...
+        $value_names = "";
+        $value_binds = "";
+        foreach ($values as $fieldname => $fieldvalue) {
+            $value_names .= ", {$fieldname}";
+            $value_binds .= ", :" . $this->addBoundVariable($fieldvalue);
+        }
+        $value_names = substr($value_names, 2);  // remove first comma
+        $value_binds = substr($value_binds, 2);  // remove first comma
+
+        return "( {$value_names} ) VALUES ( ${value_binds} )";
```
MattPrit, line 126: Should we return the value arrays rather than part of the SQL query?

api/src/Model/Services/UserData.php
```diff
@@ -70,59 +71,50 @@ private function addPersonOrProposalSearch($pid, $personId, &$args): string
...
+            $where = '1=1';
         else
             $where = 'p.login IS NOT NULL';
         $join = '';
         $extc = '';
         $group = 'GROUP BY p.personid';
-
-        if (($personId && !$isManager) || $pid || (!$isStaffMember && !$visitName))
-        {
+        $isManager = True;
```
MattPrit, line 86:
```suggestion
```
Should this be removed?

MattPrit (approved):
Looks good to me overall

api/src/Database/DatabaseQueryBuilder.php
John-Holt-Tessella: done

api/src/Model/Services/UserData.php
John-Holt-Tessella: done

MattPrit (approved):

PR #483, opened 2023-03-01
LIMS-151: Change the sort order for CHIMP scores

ndg63276:
Ticket: [LIMS-151](https://jira.diamond.ac.uk/browse/LIMS-151)

* Set "crystal" (if exists) as the default for CHIMP scores on a plate, rather than the order they come from the db

### DIFF
--- a/client/src/js/modules/shipment/views/containerplate.js
+++ b/client/src/js/modules/shipment/views/containerplate.js
@@ -655,8 +655,13 @@ define(['marionette',
         updateClasses: function() {
             var first = this.autoscores.at(0)
             var opts = ''
+            var defaultClass = "crystal"
             _.each(first.get('CLASSES'), function(prob, cl) {
-                opts += '<option value="'+cl+'">'+cl+'</option>'
+                if (cl == defaultClass) {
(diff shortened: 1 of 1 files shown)


client/src/js/modules/shipment/views/containerplate.js
```diff
@@ -655,8 +655,13 @@ define(['marionette',
         updateClasses: function() {
             var first = this.autoscores.at(0)
             var opts = ''
+            var defaultClass = "crystal"
             _.each(first.get('CLASSES'), function(prob, cl) {
-                opts += '<option value="'+cl+'">'+cl+'</option>'
+                if (cl == defaultClass) {
```
MattPrit, line 660: Is there a reason for introducing `defaultClass` rather than just comparing against `"crystal"`? Also, maybe it's a good idea to use `===` rather than `==`?

ndg63276: I felt that explicitly naming it as the defaultClass helped show why it was there.

I'll change `==` to `===`.

MattPrit: Okay, that makes sense

NKatti2011:
LGTM.

PR #491, opened 2023-03-09
LIMS-510: Display .lsa files from ANODE as preformatted logs

ndg63276:
Ticket: [LIMS-510](https://jira.diamond.ac.uk/browse/LIMS-510)

- Add .lsa to the list of extensions to be served as plain text mime type (instead of octet-stream)
- Means anode.lsa files (from ANODE) are presented as preformatted logs


ndg63276:
Hmmm... have gained extra commits, will close and try again.

PR #493, opened 2023-03-09
LIMS-217: Use pia_total_intensity as default PIA for VMXi grid scans

ndg63276:
Ticket: [LIMS-217](https://jira.diamond.ac.uk/browse/LIMS-217)

- Per-image analysis (PIA) for VMXi grid scans is stored as data collection file attachments
- So by default they are listed in attachment id order, which leaves the most useful one (pia_total_intensity) last
- By reversing the ordering, get the most useful one first, which makes it the default one displayed

### DIFF
--- a/client/src/js/modules/dc/views/gridplot.js
+++ b/client/src/js/modules/dc/views/gridplot.js
@@ -88,6 +88,7 @@ define(['jquery', 'marionette',
             this.attachments = new Attachments()
             this.attachments.queryParams.id = this.getOption('ID')
             this.attachments.queryParams.filetype = 'pia'
+            this.attachments.setSorting('DATACOLLECTIONFILEATTACHMENTID', 1)
 
             this.hasSnapshot = false
             this.snapshot = new XHRImage()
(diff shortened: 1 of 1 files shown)


client/src/js/modules/dc/views/gridplot.js
```diff
@@ -88,6 +88,7 @@ define(['jquery', 'marionette',
             this.attachments = new Attachments()
             this.attachments.queryParams.id = this.getOption('ID')
             this.attachments.queryParams.filetype = 'pia'
+            this.attachments.setSorting('DATACOLLECTIONFILEATTACHMENTID', 1)
```
John-Holt-Tessella, line 91: Does this work because pia_estimated_d_min is always created last is there a chance that it is not, in that case is it important?

John-Holt-Tessella (commented):
Try:
populatePIA: function() {
            var d = this.distl.get('data')
            if (d[0].length < 1) {
                this.ui.ty.hide()
                if (this.attachments.length) {
                    const heatMapsOptionsList = `${this.attachments.opts()} \n <option value=""> None </option>`
                    this.ui.ty2.html(heatMapsOptionsList).val("pia_estimated_d_min").show()
                    this.loadAttachment()
                }
            }
        },

ndg63276:
John's suggestion didn't quite work as the dropdown values have the datacollectionfileid's, so need to get that from the attachment list.

View both commits combined for easiest viewing.

NKatti2011:
Looks good to me. Would be good to get input from @MattPrit

MattPrit:
Looks good to me. Would be good to get input from @John-Holt-Tessella

John-Holt-Tessella (approved):

PR #494, opened 2023-03-09
LIMS-86: Allow cloning a protein with a dot in the name

ndg63276:
Ticket: [LIMS-86](https://jira.diamond.ac.uk/browse/LIMS-86)

- Add a new validation type, with dots allowed
- Add validation message for this type and a couple of others
- Add a couple of hooks for Selenium tests


MattPrit (approved):
Changes look okay; I can't think of any reason for not allowing dots in a sample's name.
@John-Holt-Tessella agrees.

PR #503, opened 2023-03-16
[LIMS-514] Fix: Change archive message on visit page

NKatti2011:
JIRA Ticket: [[LIMS-514](https://jira.diamond.ac.uk/browse/LIMS-514)]

Changes: Added condition where if the proposal is industry then the message for archival suggests deletion of data rather than archival of data. 

To test:
Check user journey is as expected.

### DIFF
--- a/client/src/js/modules/visits/views/visit_list.vue
+++ b/client/src/js/modules/visits/views/visit_list.vue
@@ -44,8 +44,9 @@
                                        title="Comment cannot be seen by User Office"
                                        v-model="visit.edited_comment"
                                        v-on:keyup.enter="onEnter(visit)" />
-                                <i data-testid="visit-table-archived" class="fa fa-archive r" title="This visit is archived, file is no longer available on disk" v-if="value.key == 'ARCHIVED' && visit.ARCHIVED == 1"></i>
-                                 
+                                <div data-testid="visit-table-archived" v-if="value.key == 'ARCHIVED' && visit.ARCHIVED == 1">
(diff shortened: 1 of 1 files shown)


client/src/js/modules/visits/views/visit_list.vue
```diff
@@ -44,8 +44,9 @@
                                        title="Comment cannot be seen by User Office"
                                        v-model="visit.edited_comment"
                                        v-on:keyup.enter="onEnter(visit)" />
-                                <i data-testid="visit-table-archived" class="fa fa-archive r" title="This visit is archived, file is no longer available on disk" v-if="value.key == 'ARCHIVED' && visit.ARCHIVED == 1"></i>
-                                 
+                                <div data-testid="visit-table-archived" v-if="value.key == 'ARCHIVED' && visit.ARCHIVED == 1">
+                                    <i class="fa fa-archive r" :title="'The raw data from this visit has been '+ isArchived + '. You can no longer reprocess data or view full sized diffraction images.'"></i>
```
ndg63276, line 48:
```suggestion
                                    <i class="fa fa-archive r" :title="'The raw data from this visit have been '+ isArchived + '. You can no longer reprocess data or view full sized diffraction images.'"></i>
```

ndg63276 (changes requested):
Sorry to be a stickler, but data is plural, so it should be "have been".

client/src/js/modules/visits/views/visit_list.vue
NKatti2011: I disagree, but I will comply.

ndg63276 (approved):

PR #511, opened 2023-03-24
Add downloading via ZipStream PHP

sshussain270:
Implemented https://github.com/maennchen/ZipStream-PHP to stream the file downloads - this doesn't write to disk or hold a lot of it in memory but rather reads and streams it.

Implemented version 2.1.0 which is compatible with PHP 7.

The downloads are now zip files instead of tar.gz files.

You will need to run `composer update` in the `/api/` directory.
And add `yum install php-zip` to Dockerfile and enable similar in production env as well.


api/tests/Page/DownloadTest.php
```diff
@@ -0,0 +1,127 @@
...
+                        $this->fail('The file is not a zip archive.');
+                        break;
+                    case ZipArchive::ER_INCONS:
+                        $this->fail('The zip archive is inconsistent.');
+                        break;
+                    case ZipArchive::ER_CRC:
+                        $this->fail('The file has a CRC error.');
+                        break;
+                    default:
+                        $this->fail("Failed to open the zip file. Zip file contents: " . $result);
```
sshussain270, line 73: It fails here with a message:
Failed to open the zip file. Zip file contents: P�H�V�[W�       file1.txt
                                                                         ��,V��̜TC=

John-Holt-Tessella (approved):

PR #513, opened 2023-03-31
Fix/fix no pdf temp dir config

John-Holt-Tessella:
If no config is set for the pdf dir the process crashes, this should not happen.


MattPrit (approved):
Looks good to me

PR #522, opened 2023-04-19
Generate image bitmaps via dials-rest server

rjgildea:
This change uses the new [dials-rest server](https://github.com/dials/dials-rest) to generate bitmaps rather than the legacy [flask-dataviewer](https://github.com/DiamondLightSource/flask-dataviewer).

Requires the variables `$dials_rest_url` and `$dials_rest_jwt` to be defined in `config.php`.


John-Holt-Tessella (approved):
Have added the original app access back in if config is not set, some better error logging for us.

PR #523, opened 2023-04-20
[LIMS-631] Fix : Undefined variables in index.php

JPHall-DLS:


John-Holt-Tessella:
PRs should be into prerelease (R1.11 has already been released to production)

PR #533, opened 2023-04-21
LIMS-634: Dont fail sample validation on fields that aren't editable

ndg63276:
Ticket: [LIMS-634](https://jira.diamond.ac.uk/browse/LIMS-634)

* Protein acronym and Sample name fields are not editable when a sample is in a container that is processing or if data has been collected on a sample
* The input validation is still looking at the hidden editable (but empty) fields, so the validation fails
* Set the validation rules to only apply to these fields if these fields are editable
* There may be a better way to do this (https://vee-validate.logaretm.com/v2/guide/conditional-and-looping-inputs.html) but I was unable to get it to work


ndg63276:
Hopefully that's a much nicer way of ensuring the validation doesn't fail.

MattPrit (approved):
Looks good to me

PR #534, opened 2023-04-21
LIMS-596: Speed up most common call to getUsers by using UNION instead of OR

ndg63276:
Ticket: [LIMS-596](https://jira.diamond.ac.uk/browse/LIMS-596)

* Calls to /api/users take ~15s due to doing 2 slow queries with `SELECT a FROM b WHERE c OR d` structure
* Much faster to use `SELECT a FROM b WHERE c UNION SELECT a FROM b WHERE d` structure
* The getUsers function has a lot of possible uses eg projects, sessions, but most common is just a list of persons on a proposal, so split those off into faster function

### DIFF
--- a/api/src/Model/Services/UserData.php
+++ b/api/src/Model/Services/UserData.php
@@ -81,6 +81,12 @@ class UserData
             $where = '1=1';
         else
             $where = 'p.login IS NOT NULL';
+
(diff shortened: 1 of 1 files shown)


sshussain270 (approved):
Looks good to me. I have tested it from multiple pages and added the report to the ticket.

PR #535, opened 2023-04-25
LIMS-612: Allow URL params in a different order

ndg63276:
Ticket: [LIMS-612](https://jira.diamond.ac.uk/browse/LIMS-612)

* URL parameters appear in order the user adds them, so /ty/ can come before or after /s/
* Add a routing option with regex with /ty/ and /s/ swapped around

This is obviously not a very generic way to achieve this, but hopefully this problem only occurs for this specific pair of swapped parameters.

### DIFF
--- a/client/src/js/modules/dc/routes.js
+++ b/client/src/js/modules/dc/routes.js
@@ -100,6 +100,21 @@ let routes = [
     }),
   },
   {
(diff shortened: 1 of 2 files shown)


client/src/js/modules/dc/routes.js
```diff
@@ -99,6 +99,21 @@ let routes = [
         sgid: +route.params.sgid || null
     }),
   },
+  {
+    path: '/dc(/visit/)?:visit([a-zA-Z]{2}[0-9]+-[0-9]+)?(/dcg/)?:dcg([0-9]+)?(/page/)?:page([0-9]+)?(/ty/)?:ty([a-zA-Z0-9_-]+)?(/s/)?:search([a-zA-z0-9_-]+)?(/id/)?:id([0-9]+)?(/pjid/)?:pjid([0-9]+)?(/sgid/)?:sgid([0-9]+)?',
```
John-Holt-Tessella, line 103: I think we need to think about this, this could easily pro;iferate. Also typing "/" in the search really confuses it. I think we need to build the url and url encode the search term

John-Holt-Tessella (approved):

PR #537, opened 2023-04-25
Pre release/fix/lims 645/make parse request paramters allow string and other types

John-Holt-Tessella:
https://jira.diamond.ac.uk/browse/LIMS-645


MattPrit (approved):
LGTM

PR #538, opened 2023-04-25
Update string match to be more like original

John-Holt-Tessella:


JPHall-DLS:
LGTM

John-Holt-Tessella:
aprroved in #537

PR #540, opened 2023-04-27
LIMS-651: Allow colons in Shipping.extra fields

ndg63276:
Ticket: [LIMS-651](https://jira.diamond.ac.uk/browse/LIMS-651)

- If one of the new Shipping.extra fields (eg scheduling restrictions) is set with a colon (eg 11:00), then the db->pq function thinks that is a variable to be substituted (eg :1, :2)
- So bind the parameters in to prevent this


John-Holt-Tessella (approved):
But check the changes i made

PR #543, opened 2023-05-03
LIMS-640: Fix visit report PDF

ndg63276:
Ticket: [LIMS-640](https://jira.diamond.ac.uk/browse/LIMS-640)

* Visit reports dont work under PHP7 strict error reporting
* FULLNAME is not a field in the database, replace with TITLE GIVENNAME FAMILYNAME

### DIFF
--- a/api/assets/pdf/visit_report.php
+++ b/api/assets/pdf/visit_report.php
@@ -27,7 +27,7 @@
         <p>Registered Users</p>
         <ul>
             <?php foreach($users as $u): ?>
-                <li><?php echo $u['FULLNAME'] ?></li>
+                <li><?php echo $u['TITLE'] ?> <?php echo $u['GIVENNAME'] ?> <?php echo $u['FAMILYNAME'] ?></li>
             <?php endforeach; ?>
         </ul>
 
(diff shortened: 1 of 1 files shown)


api/assets/pdf/visit_report.php
```diff
@@ -27,7 +27,7 @@
         <p>Registered Users</p>
         <ul>
             <?php foreach($users as $u): ?>
-                <li><?php echo $u['FULLNAME'] ?></li>
+                <li><?php echo $u['TITLE'] ?> <?php echo $u['GIVENNAME'] ?> <?php echo $u['FAMILYNAME'] ?></li>
```
John-Holt-Tessella, line 30: This comes from SQL above it and will never include TITLE so don't include it here. Does it include the next property too (LC) I think that might crash it as well

ndg63276: I believe the SQL is at https://github.com/DiamondLightSource/SynchWeb/blob/9a6f065575411e0036c0800d9ade20520f156b55/api/src/Page/PDF.php#L217 which includes TITLE. LC is set on line 205 of the same file as part of $info.

MattPrit (approved):
Looks good to me

PR #544, opened 2023-05-03
LIMS-677: Show screening method as none when strategyOption is null

ndg63276:
Ticket: [LIMS-677](https://jira.diamond.ac.uk/browse/LIMS-677)

* When a container is created by CSV upload, the DiffractionPlan.strategyOption field is often left blank/null
* Synchweb expects something in the Screening Method column to enable editing or queueing
* Show that column as 'None' when strategyOption is blank/null.


ndg63276:
Better way, just make screening method not be a required field.

PR #546, opened 2023-05-05
Dials rest

John-Holt-Tessella:
Remove index problem


John-Holt-Tessella:
https://jira.diamond.ac.uk/browse/LIMS-636

PR #553, opened 2023-05-15
LIMS-649: Increase max beam size for i23 in run stats page

ndg63276:
Ticket: [LIMS-649](https://jira.diamond.ac.uk/browse/LIMS-649)

* I23 use bigger beam sizes than other MX beamlines
* Use larger max value for x axis on beam size x/y histograms when beamline is i23, or comparing all beamlines

### DIFF
--- a/api/src/Page/Vstat.php
+++ b/api/src/Page/Vstat.php
@@ -995,7 +995,7 @@ class Vstat extends Page
             if (is_null($max) || $h['X'] > $max) $max = $h['X'];
             if (is_null($min) || $h['X'] < $min) $min = $h['X'];
         }
-        $min = is_null($min) ? 0 : floor($min/$binSize) * $binSize; // make min align with bin size, or 0 for no data
+        $min = is_null($min) ? 0 : intval(floor($min/$binSize) * $binSize); // make min align with bin size, or 0 for no data
 
         $data = array();
         foreach ($bls as $bl => $y) {
--- a/api/src/Page/Vstat.php
+++ b/api/src/Page/Vstat.php
@@ -944,10 +944,15 @@ class Vstat extends Page
         $beamlines = $this->_get_beamlines_from_type($this->ty);
         $bls = implode('\', \'', $beamlines);
 
+        $max_beam_size = 600;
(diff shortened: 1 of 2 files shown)


api/src/Page/Vstat.php
```diff
@@ -944,10 +944,15 @@ function _parameter_histogram()
         $beamlines = $this->_get_beamlines_from_type($this->ty);
         $bls = implode('\', \'', $beamlines);
 
+        $max_beam_size = 600;
```
MattPrit, line 947: Is there any particular reason for having the i23 case be the default?

api/src/Page/Vstat.php
```diff
@@ -944,10 +944,15 @@ function _parameter_histogram()
...
+        if ($this->has_arg('bl') && $this->arg('bl') != 'i23') {
```
MattPrit, line 948: Though it probably won't be needed here, it may be better to use `!==`

api/src/Page/Vstat.php
ndg63276: It was so that the "Compare all beamlines" tickbox showed the larger beamsizes.

ndg63276:
I removed the hard coded start and end values, and the references to i23, and now just use whatever data comes out.

api/src/Page/Vstat.php
```diff
@@ -987,9 +987,15 @@ function _parameter_histogram()
...
+        $min = null;
+        $max = null;
         $bls = array();
-        foreach ($hist as $h)
+        foreach ($hist as $h) {
             $bls[$h['BEAMLINENAME']] = 1;
+            if (is_null($max) || $h['X'] > $max) $max = $h['X'];
+            if (is_null($min) || $h['X'] < $min) $min = $h['X'];
+        }
+        $min = is_null($min) ? 0 : floor($min/$binSize) * $binSize; // make min align with bin size, or 0 for no data
```
ndg63276:
```suggestion
        $min = is_null($min) ? 0 : intval(floor($min/$binSize) * $binSize); // make min align with bin size, or 0 for no data
```

John-Holt-Tessella (approved):

PR #554, opened 2023-05-15
LIMS-588: Make screen name and global fields editable

ndg63276:
Ticket: [LIMS-588](https://jira.diamond.ac.uk/browse/LIMS-588)

* If a Screen is global, should be able to edit the name
* Should only be able to edit the Global status if you are in the original proposal
* Should not be able to edit Capacity as that is not a database field


sshussain270 (approved):
Looks good. Note for testing that you can only see the screen if it is Global or if you are logged into the proposal that owns it.

PR #555, opened 2023-05-15
LIMS-155: Show plate as 12x8

ndg63276:
Ticket: [LIMS-155](https://jira.diamond.ac.uk/browse/LIMS-155)

- Plates have a standard 12x8 layout, use CSS to maintain that
- Use A1-H12 names rather than 1-96 as that is what is printed on the plates

### DIFF
--- a/client/src/css/partials/_imaging.scss
+++ b/client/src/css/partials/_imaging.scss
@@ -220,3 +220,9 @@ div.img_history {
 input[name=gap] {
 	width: 30px;
 }
+
+.plate-12-wide li {
+    @mixin cols 12, 2% 0%, 0.2%;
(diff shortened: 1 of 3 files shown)


client/src/css/partials/_imaging.scss
```diff
@@ -220,3 +220,9 @@ div.img_history {
 input[name=gap] {
 	width: 30px;
 }
+
```
John-Holt-Tessella, line 223: I think we should move it here to be more specific to the imaging
I think we should centre the text
I think we should use the content-box model so it correctly calculates the sizes

John-Holt-Tessella (approved):

PR #557, opened 2023-05-16
LIMS-336: Speed up 'View Container' page

ndg63276:
Ticket: [LIMS-336](https://jira.diamond.ac.uk/browse/LIMS-336)

* When viewing a container, each sample is checked for sample groups in turn, this currently takes 0.25s per sample
* Copying the populateInitialSampleGroupValue code into the resetSamples function removes that delay (not clear why)
* For a 16 sample puck this speeds up loading by ~4s.

### DIFF
--- a/client/src/js/modules/types/mx/shipment/views/mx-container-view.vue
+++ b/client/src/js/modules/types/mx/shipment/views/mx-container-view.vue
@@ -239,6 +239,7 @@ import { ValidationObserver, ValidationProvider }  from 'vee-validate'
 
 import ContainerHistory from 'modules/shipment/collections/containerhistory'
 import Samples from 'collections/samples'
+import SampleGroups from 'collections/samplegroups'
 import Shipments from 'collections/shipments'
 import Containers from 'collections/containers'
 import Dewars from 'collections/dewars'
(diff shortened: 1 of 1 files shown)


gfrn:
The reason this works is because the changes to the sample store are no longer triggering roughly 16*14 different "input-changed" events for the pucks table, which originally caused all inputs in the table to be revalidated and rerendered, alongside any other child components.

This new method of triggering sample updates seems to allow Vue/Vuex to optimise reference/equality checks and drops the need for these renders.

gfrn (approved):

John-Holt-Tessella:
New code could everyone look and see if they are happy

client/src/js/modules/types/mx/shipment/views/mx-container-view.vue
```diff
@@ -239,6 +239,7 @@ import { ValidationObserver, ValidationProvider }  from 'vee-validate'
 
 import ContainerHistory from 'modules/shipment/collections/containerhistory'
 import Samples from 'collections/samples'
+import SampleGroups from 'collections/samplegroups'
```
ndg63276, line 242:
```suggestion
```
Can delete this import now

PR #561, opened 2023-05-17
LIMS-276: Speed up count of _get_all_containers results by removing unnecessary joins

ndg63276:
Ticket: [LIMS-276](https://jira.diamond.ac.uk/browse/LIMS-276)

* Lots of joins that might be needed depending on the `where` clause
* Only add those that are needed to an array, then implode that to get the string of actually needed joins
* `_add_to_array` function prevents duplicates in the array


John-Holt-Tessella (approved):
@ndg63276 I have made unnecessary changes to the PR but I would like to move this functionality for use in other places so I have put it in the database query builder. Could you look at this and see if you agree

ndg63276:
Yep, looks good to me!

PR #563, opened 2023-05-18
LIMS-727: Fix bug with non-staff tracking dewars on DHL

ndg63276:
Ticket: [LIMS-727](https://jira.diamond.ac.uk/browse/LIMS-727)

* Incorrect number of args in $args
* Means non-staff cannot track DHL dewars

### DIFF
--- a/api/src/Page/Shipment.php
+++ b/api/src/Page/Shipment.php
@@ -1152,7 +1152,7 @@ class Shipment extends Page
         if (!$this->has_arg('DEWARID'))
             $this->_error('No dewar specified');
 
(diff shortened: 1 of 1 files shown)


John-Holt-Tessella:
Did not see you did this I have done https://github.com/DiamondLightSource/SynchWeb/pull/565 the above doesn't solve the problem instead it makes it so that people with all_dewar will have the same problem. Are you happy to close this an approve 565 (I will dedupe the tickets).

ndg63276:
Yep, happy to abandon this one.

PR #566, opened 2023-05-18
Replace or statment with and statement for pid

sshussain270:
pid is not used anymore other than for determining access control if it is provided in users/ endpoint. Follow up ticket is https://jira.diamond.ac.uk/browse/LIMS-742

### DIFF
--- a/api/src/Model/Services/UserData.php
+++ b/api/src/Model/Services/UserData.php
@@ -103,7 +103,7 @@ class UserData
         $extc = '';
         $group = 'GROUP BY p.personid';
 
-        if (($personId && !$isManager) || $pid || (!$isStaffMember && !$visitName))
+        if ((($personId && !$isManager) || (!$isStaffMember && !$visitName)) && $pid)
         {
             $where .= $this->addPersonOrProposalSearch($pid, $currentUserId, $args);
         }
--- a/api/src/Model/Services/UserData.php
+++ b/api/src/Model/Services/UserData.php
@@ -72,7 +72,7 @@ class UserData
         return $whereClause;
     }
 
-    function getUsers($getCount, $isStaffMember, $stringMatch, $page, $sortBy = null, $proposalid = null, $personId = null, $isManager = false, $currentUserId = null, $gid = null, $sid = null, $pjid = null, $visitName = null, $perPage = 15, $isAscending = true, $isAll = false, $onlyLogins = false, $pid=null)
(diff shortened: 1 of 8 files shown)


api/src/Model/Services/UserData.php
```diff
@@ -103,7 +103,7 @@ function getUsers($getCount, $isStaffMember, $stringMatch, $page, $sortBy = null
         $extc = '';
         $group = 'GROUP BY p.personid';
 
-        if (($personId && !$isManager) || $pid || (!$isStaffMember && !$visitName))
+        if ((($personId && !$isManager) || (!$isStaffMember && !$visitName)) && $pid)
```
John-Holt-Tessella, line 106: I think this is not right look at https://github.com/DiamondLightSource/SynchWeb/blob/82b6faa413bafd1d1b535c7765a094717964a177/api/src/Page/Users.php#LL185C29-L185C32

api/src/Model/Services/UserData.php
```diff
@@ -103,7 +103,7 @@ function getUsers($getCount, $isStaffMember, $stringMatch, $page, $sortBy = null
...
         {
             $where .= $this->addPersonOrProposalSearch($pid, $currentUserId, $args);
```
John-Holt-Tessella, line 108: Instead this should be 
```suggestion
            $where .= $this->addPersonOrProposalSearch($this->proposalid, $currentUserId, $args);
```

ndg63276: When I see the error, if I log $pid and $this->proposalid, they are both blank. This may be because I am staff, but suggests this wouldn't fix the problem.

sshussain270: I think it needs to be both of your suggestiosn. If statement should check if $this->proposalid is present and if it is, then pass $this->proposalid instead of $pid because it is such an important part of query its constructing and if it isn't present, then don't attempt to construct that query in the first place. I haven't tested this new change yet.

ndg63276:
This latest `user_fix` commit has undone the work done for LIMS-596 (https://github.com/DiamondLightSource/SynchWeb/pull/534), so now going to eg /containers/add/did/62178, there is a 20s delay while waiting for the list of possible owners.

sshussain270:
> This latest `user_fix` commit has undone the work done for LIMS-596 (#534), so now going to eg /containers/add/did/62178, there is a 20s delay while waiting for the list of possible owners.

This is now fixed.

John-Holt-Tessella (approved):

PR #567, opened 2023-05-18
LIMS-397: Dont email if dewar moves from i19 to dewar store

ndg63276:
Ticket: [LIMS-397](https://jira.diamond.ac.uk/browse/LIMS-397)

Currently users get emailed about their dewar if it moves:
* from i19 to the mx dewar store
* from one beamline to another beamline (or even to the same beamline)

Users should only be emailed about their dewar if it moves:
* from an mx beamline to the dewar store
* from any beamline to a '*-storage' location

### DIFF
--- a/api/config_sample.php
+++ b/api/config_sample.php
@@ -144,6 +144,13 @@
     # and for shipment booked,
     $shipbooked_email = 'goods@server.ac.uk';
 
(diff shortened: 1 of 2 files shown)


MattPrit (approved):
Looks good to me

PR #577, opened 2023-05-31
LIMS-681: Allow editing of shipping AWB

ndg63276:
Ticket: [LIMS-681](https://jira.diamond.ac.uk/browse/LIMS-681)

* Allow editing of "Shipping Air Waybill" field of a shipment
* AWB's are numbers only [(wikipedia)](https://en.wikipedia.org/wiki/Air_waybill#:~:text=Air%20waybills%20have%20eleven%20digit%20numbers%20which%20can%20be%20used%20to%20make%20bookings%2C%20check%20the%20status%20of%20delivery%2C%20and%20a%20current%20position%20of%20the%20shipment.)


MattPrit (approved):
Looks good to me

PR #580, opened 2023-06-02
LIMS-720: Improve unit tests

gfrn:
- Some older parts of Synchweb use RequireJS for module loading, which is a package that has not been updated in the last 5 years and sees little to no active development
- RequireJS is not commonplace anymore, and with the advent of better transpilers with support for modules rather than CJS for the JS/TS ecosystem, it is not reasonable to have new pages that use RequireJS
- Jest does not plan on supporting RequireJS https://github.com/jestjs/jest/issues/17
- The older testing solution that Synchweb used, mochapack, was a weird mix of Webpack and Mocha that is also not maintained to the same degree standalone Mocha or Jest are, and in the React meetings I've had, no one seemed partial towards using mocha, so I moved to Jest
- Anything that uses RequireJS (most notably, anything with define([]) at the start of the file) cannot reasonably be tested without extensive changes to either the testing framework or the files themselves, so I believe it is not worth it to test these older files
- Views that use Vue can still be tested, however, if they depend on some older files that make use of RequireJS and make extensive use of Backbone for something older than validation, it might not work straightaway
- Therefore, the best course of action with regards for testing is creating new tests for new/newer Vue views/components (such as a few ones I've made for the changes I've made to the login page that I hope to include in a MR/PR), and not even bother with older components/views (anything with define in the first line)
- This PR includes version upgrades for Marionette/Backbone, so it's worth checking out if any page that depends heavily on them breaks because of it
TL;DR: Not worth it making tests for anything that isn't a .vue file, it only makes sense to make tests for new components or changes to existing Vue components from here on out

Since this PR builds on top of my changes to the login process for the login unit test, it depends on https://github.com/DiamondLightSource/SynchWeb/pull/562.


gfrn:
Tests *will not work* with versions of Backbone inferior to 1.2.0, or versions of Marionette below 2.3.0. 

The application breaks on Backbone 1.3.0 or higher (or Marionette 3.0.0 or higher).

PR #585, opened 2023-06-12
LIMS-779: Move archive url to config

ndg63276:
Ticket: [LIMS-779](https://jira.diamond.ac.uk/browse/LIMS-779)

* Move archiver URL to config
* Update URL parameter format
* Update function dealing with returned values
* Cope if visit ended with no beam


MattPrit (approved):
Looks good to me

John-Holt-Tessella (approved):
Are you happy with me removing the ? from the url and adding into the code and removing the unneeded package?

PR #586, opened 2023-06-12
LIMS-705: Ignore null visits

ndg63276:
Ticket: [LIMS-705](https://jira.diamond.ac.uk/browse/LIMS-705)

* New sessions with visit_number = NULL causes confusion and potential errors in dispatch request form
* So hide sessions with visit_number = NULL

### DIFF
--- a/api/src/Page/Proposal.php
+++ b/api/src/Page/Proposal.php
@@ -337,7 +337,7 @@ class Proposal extends Page
                 $p = $props[0]['ID'];
 
             $args = array($p);
-            $where = 'WHERE s.proposalid = :1';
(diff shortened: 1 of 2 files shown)


John-Holt-Tessella:
This is not just for the dispatch request form but for everything using the end point. Are you sure that this is not going to have unintended consequences elsewhere? How do I check that out?

ndg63276:
It was suggested to use the `all` parameter instead, but that means all visits from all proposals. So I have added a new parameter to explicitly ignore null visits.

John-Holt-Tessella (approved):
LGTM :-)

PR #591, opened 2023-06-19
LIMS-219: Add All Samples To Queue

ndg63276:
Ticket: [LIMS-219](https://jira.diamond.ac.uk/browse/LIMS-219)

- Add "Add All To Queue" button to GUI
- Copy queueAllSamples function to queuePageSamples, make queueAllSamples use post request
- Send in chunks of 500 as >1000 gives a PHP error
- Allow post requests

Testing instructions:

- Go to a container that is a plate, eg /containers/cid/263100
- Click 'Prepare For Data Collection'
- Click 'Add All To Queue'


John-Holt-Tessella (changes requested):
I don't think this is the right approach.

PR #595, opened 2023-06-19
LIMS-222: Apply experimental parameters to all samples

ndg63276:
Ticket: [LIMS-222](https://jira.diamond.ac.uk/browse/LIMS-222)

- Add "Apply to All" button next to "Apply to Selected" button for presets
- Link to onPresetAll function, which is the same as onPreset but for all samples

Testing instructions:

- Go to a container that is a plate, eg /containers/cid/263100
- Click 'Prepare For Data Collection'
- Add some samples to the queue
- Select a preset and click 'Add to All'

### DIFF
--- a/client/src/js/modules/imaging/views/queuecontainer.js
+++ b/client/src/js/modules/imaging/views/queuecontainer.js
@@ -600,7 +600,7 @@ define(['marionette',
             e.preventDefault()
 
             var p = this.plans.findWhere({ DIFFRACTIONPLANID: this.ui.preset.val() })
-            if (p) this.applyModelAll(p, true)
+            if (p) this.applyModel(p, true)
         },
 
         applyModel: function(p, all) {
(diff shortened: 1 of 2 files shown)


ndg63276:
I don't think this is a problem. Users can already choose a large page size, tick them all, and then apply a preset to all on the page.

client/src/js/modules/imaging/views/queuecontainer.js
```diff
@@ -592,11 +593,22 @@ define(['marionette',
...
-        applyModel: function(p) {
-            var models = this.qsubsamples.where({ isGridSelected: true })
+        applyPresetAll:function(e) {
+            e.preventDefault()
+
+            var p = this.plans.findWhere({ DIFFRACTIONPLANID: this.ui.preset.val() })
+            if (p) this.applyModel(p, true)
+        },
+
+        applyModel: function(p, all) {
```
John-Holt-Tessella, line 606: p and all are both a bit ambigous lets make it:  
  p - modelParamaeters 
  all - isLimitedToSelected
This way I think it is easier to understand the code.

client/src/js/modules/imaging/views/queuecontainer.js
```diff
@@ -592,11 +593,22 @@ define(['marionette',
...
+            if (all) {
+                var models = this.qsubsamples.fullCollection.filter(function(ss) { return true })
```
John-Holt-Tessella, line 608: This doesn't need a filter use a convert to array

PR #597, opened 2023-06-19
LIMS-59: Never write null to DewarTransportHistory.storageLocation

ndg63276:
Ticket: [LIMS-59](https://jira.diamond.ac.uk/browse/LIMS-59)

* DewarTransportHistory.storageLocation should not be nullable in the schema
* Write the dewar location as "off-site" when creating an AWB, booking a pickup or marking as sent
* Write the dewar location as the current location when cancelling a pickup, in case dewar has moved in the meantime
* Do the same for the dewar history

### DIFF
--- a/api/src/Model/Services/AssignData.php
+++ b/api/src/Model/Services/AssignData.php
@@ -90,10 +90,10 @@ class AssignData
         $location = $this->db->pq("SELECT storagelocation FROM dewar WHERE dewarid=:1", array($dewarId));
         $this->updateDewarHistory($dewarId, 'unprocessing', $location[0]['STORAGELOCATION']);
 
-        $conts = $this->db->pq("SELECT containerid as id FROM container WHERE dewarid=:1", array($dewarId));
+        $conts = $this->db->pq("SELECT containerid FROM container WHERE dewarid=:1", array($dewarId));
         foreach ($conts as $container)
         {
-            $this->updateContainerAndHistory($container['ID'], 'at facility', '', '');
+            $this->updateContainerAndHistory($container['CONTAINERID'], 'at facility', '', '');
         }
     }
 
--- a/api/src/Model/Services/AssignData.php
+++ b/api/src/Model/Services/AssignData.php
@@ -90,10 +90,10 @@ class AssignData
         $location = $this->db->pq("SELECT storagelocation FROM dewar WHERE dewarid=:1", array($dewarId));
         $this->updateDewarHistory($dewarId, 'unprocessing', $location[0]['STORAGELOCATION']);
 
-        $conts = $this->db->pq("SELECT containerid FROM container WHERE dewarid=:1", array($dewarId));
+        $conts = $this->db->pq("SELECT containerid as id FROM container WHERE dewarid=:1", array($dewarId));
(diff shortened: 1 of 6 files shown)


api/src/Model/Services/AssignData.php
```diff
@@ -85,9 +85,10 @@ function updateDewar($dewarId, $status)
 
     function deactivateDewar($dewarId)
     {
-        $this->updateDewarHistory($dewarId, 'unprocessing');
+        $location = $this->db->pq("SELECT storagelocation FROM dewar WHERE dewarid=:1", array($dewarId));
```
John-Holt-Tessella, line 88: I think that dewar.storagelocation can be null, maybe do a check here and set it to blank if it is null?

John-Holt-Tessella: Also line 105 look like it could be null if the beamline is not set on the container

api/src/Page/Shipment.php
```diff
@@ -3090,7 +3090,7 @@ function _cancel_pickup()
             $this->_error('No such lab contact');
         $cont = $cont[0];
 
-        $dewars = $this->db->pq("SELECT d.dewarid
+        $dewars = $this->db->pq("SELECT d.dewarid, d.storagelocation
```
John-Holt-Tessella, line 3093: could the dewar storage location not be null?

John-Holt-Tessella (changes requested):
I think there are several other places that this value could be null. Would it be wise to add a single function for adding dewar history? Maybe by using the updateDewarHistory funtion in assign data? Although it is a bit of a weird name for it. Feels like it would be better in a dewarData function?

api/src/Model/Services/AssignData.php
ndg63276: I've added a check just before the insert statement (line 105)

api/src/Page/Shipment.php
```diff
@@ -1562,16 +1562,16 @@ function _send_shipment()
         }
 
         $this->db->pq("UPDATE shipping SET shippingstatus='sent to facility' where shippingid=:1", array($ship['SHIPPINGID']));
-        $this->db->pq("UPDATE dewar SET dewarstatus='sent to facility' where shippingid=:1", array($ship['SHIPPINGID']));
+        $this->db->pq("UPDATE dewar SET dewarstatus='sent to facility', storagelocation='off-site' where shippingid=:1", array($ship['SHIPPINGID']));
```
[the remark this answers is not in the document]
ndg63276: I've added a check for null on line 1054

api/src/Page/Shipment.php
ndg63276: I've added a check for null on line 3116

ndg63276:
Refactoring into a single function is probably a good idea, but maybe that should be a new ticket. Hopefully have resolved the issues with any insert into DTH statements.

John-Holt-Tessella (approved):

John-Holt-Tessella (changes requested):
Tests do't pass

ndg63276:
> Tests don't pass

Tests now pass :)

PR #603, opened 2023-06-22
LIMS-210: Prevent multiple samples being created in the same location…

ndg63276:
… in the same container

Ticket: [LIMS-210](https://jira.diamond.ac.uk/browse/LIMS-210)

If Synchweb attempts to create multiple samples in the same container and the same location, this is probably an error and should be ignored.
Added a check for sublocation for future-proofing.

### DIFF
--- a/api/src/Page/Sample.php
+++ b/api/src/Page/Sample.php
@@ -1319,7 +1319,18 @@ class Sample extends Page
         if ($this->has_arg('collection')) {
             $this->db->start_transaction();
             $col = array();
+            $used = array();
             foreach ($this->arg('collection') as $s) {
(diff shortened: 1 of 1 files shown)


John-Holt-Tessella:
Weirdly I am getting no sample 10 populated and the error:
[31-Aug-2023 14:38:31] WARNING: [pool www] child 2209 said into stderr: "NOTICE: PHP message: Container location already in use: 271089.10"

I am also not sure I like throwing away user input like this, would it be possible to get a message back to the user saying that the sample location X has been duplicated and therefore it should be double checked.

ndg63276:
I guess  271089.10 ==  271089.1, I should use a different delimiter, will have a look tomorrow.

PR #606, opened 2023-06-28
LIMS-116: Summary page revamp

NKatti2011:
Ticket: [LIMS-116](https://jira.diamond.ac.uk/browse/LIMS-116)

  The current summary page currently displays information that is not completely relevant such as exposure time/oscillation. The page revamp involves adding in more useful information and the ability to filter across these variables. The current changes are only visible to staff and if the config is configuredThe changes included are listed below:

- Add page consisting of columns of requested relevant data including but not limited to overall resIOverSigI2, inner rMeasWithinIPlusIMinus, and outer resolutionLimitHigh. 
- Add ability to remove and add columns in the table for results personalisation
- Add ability to filter results based on column. Filters include ability to search values greater than, less than, equal to, or like a value that you input. 
- Add grouped results in rows which then expand to reveal further results on row click. Results are grouped by data collection.
- IFrame introduced as a preview to data collections page. 

Testing:

  Log in 
 Navigate to summary page
 Check results load


api/src/TemplateParser.php
```diff
@@ -20,6 +20,8 @@ class TemplateParser
         var $params = array();
 
         function __construct($db) {
+
+            print "In BaseClass constructor\n";
```
gfrn, line 24:
```suggestion
```

api/src/Database/Type/PureMySQL.php
```diff
@@ -0,0 +1,262 @@
+<?php
```
gfrn, line 1: This file is very similar to MySQL.php, is there a way to extend MySQL.php with what you need?

client/src/js/modules/summary/views/summary.vue
```diff
@@ -0,0 +1,1775 @@
...
+                this.pillIndex = 4;
+            }
+
+            if (this.windowWidth > 2200 ) {
+                this.pillIndex = 5;
+            }
+
+            if (this.windowWidth > 2800 ) {
+                this.pillIndex = 6;
+            }
```
MattPrit, line 1316: Could we reverse the order of these and make them else-ifs?

gfrn: You can fit these to a function if you make the values more homogeneous, or possibly even use CSS to do this

client/src/js/modules/summary/views/summary.vue
```diff
@@ -0,0 +1,1775 @@
...
+                    return [];
+                };
+        },
+        checkedColumns(index) {
+            // add columns to selected or deselected object for pill display
+            this.summaryParameters[index].checked = !this.summaryParameters[index].checked;
+
+            if (this.summaryParameters[index].checked == false){
+
+            this.deselectedColumns.push(this.summaryParameters[index]);
```
MattPrit, line 1280: Do we need these two arrays? Could in stead each of `summaryParameters` have a `selected` attribute?

client/src/js/modules/summary/views/summary.vue
```diff
@@ -0,0 +1,1775 @@
...
+        },
+        togglePills(value) {
+            // remove selected columns using pills x icon
+            for (var index in this.summaryParameters) {
+
+                if (this.summaryParameters[index].title == value) {
+                    this.summaryParameters[index].checked = false;
+                    this.deselectedColumns.push(this.summaryParameters[index]);
+                    this.selectedColumns = this.selectedColumns.filter(item => item !== this.summaryParameters[index]);
+                }
```
MattPrit, line 1328: Is this duplicated logic from `checkedColumns`?

MattPrit:
A couple of extra questions:
- Can we remove the proposal dropdown?
- Can we make the iframe render only the data collection component rather than the whole of SynchWeb?

api/src/Database/DatabaseFactory.php
```diff
@@ -19,25 +20,28 @@ function __construct($databaseConnectionFactory)
...
-            $dbClassName = $this->database_types[strtolower($database_type)];
+        if (key_exists(strtolower($databaseType), $this->database_types)) {
+            $dbType = $this->database_types[strtolower($databaseType)];
 
-            $full_class_name = 'SynchWeb\\Database\\Type\\' . $dbClassName;
+            $full_class_name = 'SynchWeb\\Database\\Type\\' . $dbType["dbClassName"];;
 
             if (class_exists($full_class_name)) {
-                $conn = $this->databaseConnectionFactory->get($dbClassName);
+                error_log($dbType["dataConnectionName"]);
```
John-Holt-Tessella, line 43:
```suggestion
```

api/src/Database/DatabaseConnectionFactory.php
```diff
@@ -24,6 +25,16 @@ public function get($databaseType)
             $conn = new \mysqli($host, $isb['user'], $isb['pass'], $dbn, $port);
             $conn->set_charset("utf8mb4");
         }
+        if ($databaseType == 'PureMySQL') {
+            $port = array_key_exists('port', $summarydbconfig) ? $summarydbconfig['port'] : null;
```
John-Holt-Tessella, line 29: Can we have a think about what to do if the facility is not running a summary database, without the $summarydbconfig set this code dies. We probably need to carry on and do nothing and make sure the page is fine if we get a null back.

client/src/js/app/components/navbar.vue
```diff
@@ -10,6 +10,17 @@
     class="tw-hidden tw-z-10 md:tw-flex md:flex-row tw-bg-sidebar-grad-end tw-justify-center tw-my-4 md:tw-mx-auto tw-border tw-border-gray-400 tw-divide-x tw-divide-gray-400"
   >
     <!-- List proposals item -->
+    <keep-alive>
+      <router-link
+        v-if="isStaff"
```
John-Holt-Tessella, line 15: Could I have the option of turning this off from the config, so other facilities that don't have this database setup don't have this link

api/src/Database/Type/PureMySQL.php
NKatti2011: PureMySQL removes functionality from MySQL, so extending it will not be useful.

api/src/TemplateParser.php
NKatti2011: Confused as not part of my commit but will commit anyway..

client/src/js/app/components/expandable-sidebar.vue
```diff
@@ -0,0 +1,112 @@
...
+    },
+    methods: {
+        toggleSidebar() {
+        this.isOpen = !this.isOpen;
+
+        this.contentDelay()
+        },
+        contentDelay() {
+ 
+            if (!this.loadContent) {
```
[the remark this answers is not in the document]
NKatti2011: its mimicking a css transition where display is toggled with a transition - not easily done just with css.

client/src/js/app/components/custom-accordian.vue
```diff
@@ -0,0 +1,83 @@
...
+    methods: {
+      toggleAccordion() {
+        this.isOpen = !this.isOpen;
+
+        this.contentDelay()
+      },
+
+      contentDelay() {
+
+        if (!this.loadContent) {
```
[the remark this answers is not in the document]
NKatti2011: custom-accordian is no longer used so deleted the entire file.

client/src/js/modules/summary/views/summary.vue
```diff
@@ -0,0 +1,1775 @@
...
+            deep: true
+        },
+    }
+
+    
+}
+
+</script>
+
+<style>
```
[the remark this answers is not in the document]
NKatti2011: I split the styling into summary.scss file in css/partials.

client/src/js/modules/summary/views/summary.vue
NKatti2011: Changed to this approach which seems to work and is cleaner:

```
 if (this.windowWidth > 1000 & this.windowWidth < 2900) {

                const windWidthRounded = Math.ceil(this.windowWidth/100)*100;

                if( windWidthRounded % 200 === 0) {

                    this.pillIndex = (windWidthRounded - 1000)/200;

                }

            } 
```

api/src/Database/Type/PureMySQL.php
NKatti2011: After discussions, will leave as is as it removes oracle functionality which may be useful for new pages/endpoints in future.

PR #607, opened 2023-06-29
LIMS-844: SSO-specific URL

gfrn:
Ticket: [LIMS-844](https://jira.diamond.ac.uk/browse/LIMS-844)

- Fixes bug caused by services that use CAS (as in CAS, the protocol)
- Allows multiple authentication mechanisms to coexist side by side in the codebase
- Sends client details through Authorization header in OIDC token endpoint.

### DIFF
--- a/api/src/Authentication/Type/OIDC.php
+++ b/api/src/Authentication/Type/OIDC.php
@@ -60,7 +60,7 @@ class OIDC extends AuthenticationParent implements AuthenticationInterface
     {
         global $cookie_key;
 
-        if (array_key_exists("cookie_key", $_COOKIE)) {
+        if (array_key_exists($cookie_key, $_COOKIE)) {
             return($this->getUser($_COOKIE[$cookie_key]));
         }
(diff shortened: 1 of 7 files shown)


api/src/Authentication/Type/OIDC.php
```diff
@@ -56,7 +59,12 @@ function authenticate($login, $password)
     function check()
     {
         global $cookie_key;
-        return($this->getUser($_COOKIE[$cookie_key]));
+
+        if (array_key_exists("cookie_key", $_COOKIE)) {
```
John-Holt-Tessella, line 63:
```suggestion
        if (array_key_exists($cookie_key, $_COOKIE)) {
```

gfrn: As per PHP docs, the key is a string or an int:

https://www.php.net/manual/en/function.array-key-exists.php

` array_key_exists(string|int $key, array $array): bool`

PR #609, opened 2023-06-30
LIMS-774 - Display single crystal vs multi crystal data in appropriate templates

gfrn:
Ticket: [LIMS-774](https://jira.diamond.ac.uk/browse/LIMS-774)

### DIFF
--- a/api/src/Page/DC.php
+++ b/api/src/Page/DC.php
@@ -153,6 +153,10 @@ class DC extends Page
                         LEFT OUTER JOIN processingjob pj ON dc.datacollectionid = pj.datacollectionid
                         LEFT OUTER JOIN autoprocprogram app ON (app.autoprocprogramid = api.autoprocprogramid OR dc.datacollectionid = pj.datacollectionid)
                         INNER JOIN autoprocprogrammessage appm ON appm.autoprocprogramid = app.autoprocprogramid";
+            } else if ($this->arg('t') == "scrystal" || $this->arg('t') == "nscrystal") {
+                // Single crystal or explicitly non-single-crystal fields
+                $where = ($this->arg('t') == "nscrystal") ? ' AND NOT ' : ' AND ';
+                $where .= '(dcg.experimentType IS NOT NULL AND dcg.experimentType in ("OSC", "Diamond Anvil High Pressure"))';
             }
         }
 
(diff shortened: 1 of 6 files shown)


api/src/Page/DC.php
```diff
@@ -986,8 +990,8 @@ function _data_collections($single = null)
 
                 if ($dc['DCT'] == 'Mesh')
                     $dc['DCT'] = 'Grid Scan';
-                if ($dc['DCT'] == 'OSC')
-                    $dc['DCT'] = 'Data Collection';
+                if ($dc['DCT'] == 'OSC' || $dc['DCT'] == 'Diamond Anvil High Pressure') 
+                    $dc['DCT'] = 'Single Crystal';
```
John-Holt-Tessella, line 994: Because you are no longer setting OSC to Data Collection I think you need to add Single Crystal to dc.html line 29. So it works for non xpdf.

Also how would you feel about not doing this change to single crystal but passing either OSC or Diamond Anvil High Pressure to the front end then changing `collectionTypeSpecific: { "Single Crystal": MXView },` to `collectionTypeSpecific: { "Diamond Anvil High Pressure": MXView ,
                                         "OSC": MXView },

John-Holt-Tessella (commented):
Are my comments fair?
But it looks really good, 3 days was a bit of an over exaggeration I think

api/src/Page/DC.php
gfrn: That makes sense, it's a good bit of futureproofing

api/src/Page/DC.php
```diff
@@ -153,6 +153,10 @@ function _data_collections($single = null)
                         LEFT OUTER JOIN processingjob pj ON dc.datacollectionid = pj.datacollectionid
                         LEFT OUTER JOIN autoprocprogram app ON (app.autoprocprogramid = api.autoprocprogramid OR dc.datacollectionid = pj.datacollectionid)
                         INNER JOIN autoprocprogrammessage appm ON appm.autoprocprogramid = app.autoprocprogramid";
+            } else if ($this->arg('t') == "scrystal" || $this->arg('t') == "nscrystal") {
+                // Single crystal or explicitly non-single-crystal fields
+                $where = ($this->arg('t') == "nscrystal") ? ' AND NOT ' : ' AND ';
+                $where .= '(dcg.experimentType IS NOT NULL AND dcg.experimentType in ("OSC", "Diamond Anvil High Pressure"))';
```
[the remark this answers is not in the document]
gfrn: I left it in just in case, but I guess from now onwards nothing will be inserted as NULL

api/src/Page/DC.php
```diff
@@ -156,7 +156,8 @@ function _data_collections($single = null)
             } else if ($this->arg('t') == "scrystal" || $this->arg('t') == "nscrystal") {
                 // Single crystal or explicitly non-single-crystal fields
                 $where = ($this->arg('t') == "nscrystal") ? ' AND NOT ' : ' AND ';
-                $where .= '(dcg.experimentType IS NOT NULL AND dcg.experimentType in ("OSC", "Diamond Anvil High Pressure"))';
+                // This IS NOT NULL is not redundant; this condition always evalutes to TRUE with AND NOT without it
+                $where .= 'dcg.experimentType in ("OSC", "Diamond Anvil High Pressure")';
```
John-Holt-Tessella, line 160:
```suggestion
                $where .= '(dcg.experimentType IS NOT NULL AND dcg.experimentType in ("OSC", "Diamond Anvil High Pressure"))';
```

John-Holt-Tessella (approved):

PR #610, opened 2023-07-03
Make general dc path less important

John-Holt-Tessella:
The path for DC page was made general so that it could accept arguments in any order but having done this it now takes the path from other routes. Move it doen the routing table so it matches last instead of first.

### DIFF
--- a/client/src/js/modules/dc/router.js
+++ b/client/src/js/modules/dc/router.js
@@ -4,7 +4,6 @@ define(['utils/lazyrouter'], function(LazyRouter) {
     var Router = LazyRouter.extend({
         appRoutes: {
             'dc': 'dc_list',
-            
             'dc/view/id/:id': 'di_viewer',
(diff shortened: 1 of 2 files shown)


gfrn:
Works on my local deployment, however, we might need to make the path regular expression less generic so this isn't a problem in the future (maybe even write some tests for it!)

client/src/js/modules/dc/router.js
```diff
@@ -4,14 +4,15 @@ define(['utils/lazyrouter'], function(LazyRouter) {
     var Router = LazyRouter.extend({
         appRoutes: {
             'dc': 'dc_list',
-            'dc(/visit/:visit)(:pathmatch)': 'dc_list',
+
```
gfrn, line 7:
```suggestion
```

John-Holt-Tessella:
OK, will create ticket to write tests, not sure that the reg ex can be less because it is matching every segment of the path which some parts can have any values

PR #621, opened 2023-07-06
LIMS-95: Show max score for each drop in a plate

ndg63276:
Ticket: [LIMS-95](https://jira.diamond.ac.uk/browse/LIMS-95)

* Once a plate is imaged, the previous drop scoring is hidden, so a user may have to click through each drop again and score them
* Add a checkbox to show the 'max' score for each drop, which will allow them to quickly see which drops have been highly scored in the past

Testing (on dev db):
* go to proposal 30330, go to a container
* click on some drops, set a score, change inspections and repeat
* tick 'Max scores' radio button to show the highest score colour for each drop
* Also try "Current scores", "Data Status" and "Auto Scores" with various drop downs (Schema CHIMP should not be visible)

### DIFF
--- a/api/src/Page/Imaging.php
+++ b/api/src/Page/Imaging.php
@@ -557,7 +557,7 @@ class Imaging extends Page
 
 
 
-    # Get a list of sample images avaialble for an inspection
+    # Get a list of sample images available for an inspection
     function _get_inspection_images()
     {
         $where = '';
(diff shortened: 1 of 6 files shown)


ndg63276:
LGTM.

John-Holt-Tessella (approved):
The backend part is good. @ndg63276 has reviewed the part I did and is happy

PR #622, opened 2023-07-10
LIMS-77: Link Lab Contacts to fedids

ndg63276:
Ticket: [LIMS-77](https://jira.diamond.ac.uk/browse/LIMS-77)

* Give users a new dropdown when creating a lab contact, to select either their own login, or blank for someone else
* Prefill the form if we have the information
* Dont allow them to create a lab contact linked to someone else's fedid
* Only allow one personId/proposalId combination
* Get actual postcode instead of blank string


client/src/js/modules/contact/views/addcontact.js
```diff
@@ -17,26 +18,68 @@ define(['views/form',
...
+            }
         },
 
         initialize: function() {
             this.countries = new Countries()
             this.countries.state.pageSize = 9999
             this.ready = this.countries.fetch()
+            this.users = new Users()
+            this.users.queryParams.login = 1
+            this.ready2 = this.users.fetch()
```
gfrn, line 59:
```suggestion
            this.countries = new Countries()
            this.countries.state.pageSize = 9999
            this.users = new Users()
```

Used this opportunity to do a bit of rewriting; `this.ready` is redundant, and we don't really need the `queryParams`, seems like the API responds just fine without it (unless you know of any weird edge case)

client/src/js/modules/contact/views/addcontact.js
```diff
@@ -17,26 +18,68 @@ define(['views/form',
...
         },
 
         createModel: function() {
             this.model = new Contact()
         },
-        
+
         onRender: function() {
             $.when(this.ready).done(this.populateCountries.bind(this))
+            $.when(this.ready2).done(this.populateUsers.bind(this))
```
gfrn, line 68:
```suggestion
            Promise.all([this.countries.fetch(), this.users.fetch()]).then(() => {
                this.populateCountries();
                this.populateUsers();
            })
```

With `Promise.all`, we can do concurrent fetches without having to resort to bound functions or weird jQuery shenanigans, since the lexical scope of the anonymous function provided in the body of the thenable is always going to be the state of the global `this` object at that given moment in time

client/src/js/modules/contact/views/addcontact.js
ndg63276: The queryParams.login is to only show users with a fedid field, see https://github.com/DiamondLightSource/SynchWeb/blob/master/api/src/Controllers/UserController.php#L193 and https://github.com/DiamondLightSource/SynchWeb/blob/master/api/src/Model/Services/UserData.php#L92

gfrn: Oh, I see, I thought login mapped directly to the login column instead, I'll edit the suggestion

gfrn (approved):

PR #624, opened 2023-07-11
Fix routing em datacollection

John-Holt-Tessella:
https://jira.diamond.ac.uk/browse/LIMS-872

Routing is in wrong order because em dc routing is captured by dc general pattern first


ndg63276 (approved):
Works for me

PR #626, opened 2023-07-11
LIMS-871: Fix wrong number of dewar uses for staff

ndg63276:
Ticket: [LIMS-871](https://jira.diamond.ac.uk/browse/LIMS-871)

* Staff see wrong number of uses of each dewar on the Registered Dewars page, they see (actual number of uses) * (number of proposals the dewar belongs to)
* Staff also get redirected to the max(proposal) when clicking on a registered dewar, and then only see the shipments for that dewar on that proposal

Testing:
1.  Log in as a staff member, go to proposal mx29907.
2. Go to Registered Dewars page
3. Search for DLS-MX-0873, it should show "# Uses" as around 19 (not 57)
4. Click on DLS-MX-0873, proposal should stay as mx29907 (not switch to mx30951)
5. The shipments section should show all 19 uses of this dewar, even those that belong to other proposals
6. Repeat as a user from mx29907, the only difference should be in step 5, it should only show shipments from mx29907

### DIFF
--- a/api/src/Page/Shipment.php
+++ b/api/src/Page/Shipment.php
@@ -568,7 +568,7 @@ class Shipment extends Page
         $args = array($this->proposalid);
         $where = 'p.proposalid=:1';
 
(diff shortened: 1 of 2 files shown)


MattPrit (approved):
Looks good to me

PR #627, opened 2023-07-11
Fix/lims 876/check unset variables

John-Holt-Tessella:
https://jira.diamond.ac.uk/browse/LIMS-876


ndg63276:
Code looks ok to me, although I haven't tested it (not sure how to use OIDC on a local dev server).

PR #629, opened 2023-07-14
LIMS-892: Dont allow UDC samples with no screening method and no reqd resolution

ndg63276:
Ticket: [LIMS-892](https://jira.diamond.ac.uk/browse/LIMS-892)

https://github.com/DiamondLightSource/SynchWeb/pull/558 (LIMS-677) introduced a bug where users could leave screening method and required resolution both blank and then queue for UDC. Required resolution should be a required field if queueing for UDC, and even if screening method is blank. (The extra comma in the rules line matches when screening method is blank).


MattPrit (approved):
Looks good to me

PR #633, opened 2023-07-18
LIMS-898: Fix callouts in visit stats

ndg63276:
Ticket: [LIMS-898](https://jira.diamond.ac.uk/browse/LIMS-898)

* Main bug was returning the object containing the callouts instead of an array of the callouts (foreach clause on $calls_tmp)
* Moved URLs to config.php
* Removed the blank columns from callouts
* Fixed callouts and elog entries divs being flipped

### DIFF
--- a/api/config_sample.php
+++ b/api/config_sample.php
@@ -62,6 +62,12 @@
     # URL to access the PV archiver
     $archive_url = '';
 
+    # URL to access elog logbook
+    $elog_base_url = '';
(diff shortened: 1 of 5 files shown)


John-Holt-Tessella (approved):
I can not run this but the code looks fine.
Could you update the configs (or make sure the configs are updated) now so we don't forget; The config change will not effect the running services.

PR #648, opened 2023-08-23
LIMS-759: Show multipin samples and their sublocations

ndg63276:
**JIRA ticket**: [LIMS-759](https://jira.diamond.ac.uk/browse/LIMS-759)

**Summary**:

Multipin samples use the BLSample.subLocation field to put multiple samples in the same container/location. These samples should be displayed and editable in Synchweb.

**Changes**:
- Return sublocation from the database
- Make space for up to 10 sublocations per sample in the state.sample array, by using (10*location)+sublocation as the index.
- Use this index to tag each row for editing, reqired fields etc, rather than the location
- Add a sublocation column in the GUI
- Show empty locations but not empty sublocations

**To test**:
- Test samples in sublocations are shown on eg /containers/cid/270597
- Test editing samples individually and with the bulk editor
- Test containers without sublocations still look ok eg /containers/cid/245858


gfrn (approved):

NKatti2011:
Rolling back as change has broken puck creation

PR #652, opened 2023-08-25
LIMS-954: Add ability to edit manufacturer serial no on dewars

ndg63276:
**JIRA ticket**: [LIMS-954](https://jira.diamond.ac.uk/browse/LIMS-954)

**Summary**:

Now that the DewarRegistry table has a manufacturerSerialNumber column, we should make it editable in Synchweb.

**Changes**:
- Get serial number from database
- Update the INSERT statement for new dewars to allow serial number
- Update the loop that does the UPDATE statement for editing existing dewars
- Add optional text fields to both the new dewar and existing dewar pages
- Clear the field once a new dewar has been created

**To test**:
- Add a new Registered Dewar with and without a sserial number
- Edit an existing dewar to add a serial number (see PR https://github.com/DiamondLightSource/SynchWeb/pull/651 if you find a bug)

### DIFF
--- a/api/src/Page/Shipment.php
+++ b/api/src/Page/Shipment.php
@@ -53,6 +53,7 @@ class Shipment extends Page
         'STATUS' => '[\w|\s|\-]+',
 
         'PURCHASEDATE' => '\d+-\d+-\d+',
+        'MANUFACTURERSERIALNUMBER' => '.*',
         'LABCONTACTID' => '\d+',
         'REPORT' => '.*',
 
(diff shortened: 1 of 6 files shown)


gfrn:
Going to a registered dewar's page (https://ispyb-test.diamond.ac.uk/dewars/registry/DLS-MX-3333) pops up an error saying "No such model", otherwise everything works fine

ndg63276:
The direct links don't work as no proposal is set. I could fix it for staff as they should be able to see all dewars, but not for non-staff, as they need the proposal to check the permissions. Not sure the gain is worth it, and probably beyond the scope of this ticket.

gfrn:
> The direct links don't work as no proposal is set. I could fix it for staff as they should be able to see all dewars, but not for non-staff, as they need the proposal to check the permissions. Not sure the gain is worth it, and probably beyond the scope of this ticket.

That's alright! I don't think it matters much, this was mostly to check out the "Edit an existing dewar to add a serial number" test requirement, so if that is not a concern anyway, I'll approve it

gfrn (approved):

PR #660, opened 2023-09-04
LIMS-210: Prevent double clicking to create multiple samples

ndg63276:
Ticket: [LIMS-210](https://jira.diamond.ac.uk/browse/LIMS-210)

Changes:
* We now know that double-clicking on the tick button when editing samples can create 2 samples in the same location. By closing the editor before saving, we prevent that.

Tests:
* Editing a sample and single clicking creates one sample
* Editing a sample and double clicking only creates one sample
* Editing a sample, leaving invalid parameters, clicking should leave the editor open and not create a sample

### DIFF
--- a/client/src/js/modules/types/mx/shipment/views/container-mixin.js
+++ b/client/src/js/modules/types/mx/shipment/views/container-mixin.js
@@ -321,11 +321,10 @@ export default {
       try {
         this.$store.commit('loading', true)
         const result = await this.$refs.containerForm.validate()
-
         if (result) {
-          await this.saveSample(location)
           const samplesRef = this.$refs.samples
           samplesRef.$refs[`sample-row-${location}`][0].closeSampleEditing()
+          await this.saveSample(location)
           this.$refs.containerForm.reset()
         }
         else {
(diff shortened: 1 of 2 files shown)


client/src/js/modules/types/mx/shipment/views/container-mixin.js
```diff
@@ -321,11 +321,10 @@ export default {
       try {
         this.$store.commit('loading', true)
         const result = await this.$refs.containerForm.validate()
-
         if (result) {
-          await this.saveSample(location)
           const samplesRef = this.$refs.samples
           samplesRef.$refs[`sample-row-${location}`][0].closeSampleEditing()
+          await this.saveSample(location)
```
John-Holt-Tessella, line 327: There are two issues here:
1. If the sample can not be saved then the edit row has been closed already
2. It is possible that the await in 323 gives the user a chance to do a second click
I think we should use the loading status to avoid this.

John-Holt-Tessella: See latest commit

ndg63276: Yep, that works for me.

PR #661, opened 2023-09-07
LIMS-997: Don't allow incoming shipments on in/sw proposals

ndg63276:
JIRA ticket: [LIMS-997](https://jira.diamond.ac.uk/browse/lims-997)

Description:

When creating an air waybill, hide the "Use Facility Account" button for in/sw proposals. This is already done for return shipments (see https://github.com/DiamondLightSource/SynchWeb/pull/393).

Changes:
* Only show "Use Facility Account" button if proposal doesn't start with in or sw.

To test:
* Test cannot use facility account for industrial proposals
* Test can still use facility account for non-industrial proposals


ndg63276:
Closing as sometimes in/sw should be allowed to ship (eg CRUK)

PR #662, opened 2023-09-11
LIMS-960: Add logging to SynchWeb's OIDC auth

gfrn:
JIRA ticket: [LIMS-960](https://jira.diamond.ac.uk/browse/LIMS-960)

Changes:
- Add log messages for discovery endpoint connection failures

### DIFF
--- a/api/src/Authentication/Type/OIDC.php
+++ b/api/src/Authentication/Type/OIDC.php
@@ -18,6 +18,12 @@ class OIDC extends AuthenticationParent implements AuthenticationInterface
         curl_setopt($ch, CURLOPT_HEADER, 0);
         curl_setopt($ch, CURLOPT_RETURNTRANSFER, 1);
         $response = curl_exec($ch);
+
+        if (curl_errno($ch) || curl_getinfo($ch, CURLINFO_HTTP_CODE != 200)) {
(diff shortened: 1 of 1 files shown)


api/src/Authentication/Type/OIDC.php
```diff
@@ -18,14 +18,20 @@ function __construct() {
         curl_setopt($ch, CURLOPT_HEADER, 0);
         curl_setopt($ch, CURLOPT_RETURNTRANSFER, 1);
         $response = curl_exec($ch);
+
+        if (curl_errno($ch) || curl_getinfo($ch, CURLINFO_HTTP_CODE != 200)) {
+            error_log("Failed to connect to OIDC discovery endpoint");
```
John-Holt-Tessella, line 23: would it be worth logging the errno or http code at this point too?

gfrn: Since this is an open, relatively simple endpoint, I believe errors will amount to either 404s, 500s or timeouts (the latter being the most likely), so I think it's not as pivotal in this instance. However, since it's easy to add and there are no drawbacks, I'll add it

John-Holt-Tessella (approved):

PR #674, opened 2023-09-25
LIMS-1027: Allow emails to be sent whwn LDAP server is not configured

John-Holt-Tessella:
**JIRA ticket**:  [TICKET-1027](https://jira.diamond.ac.uk/browse/LIMS-1027)

**Summary**:

If in config.php ldap server is set to null then no name lookup is done when emails are sent. This allows us to test sending email in dev 

**Changes**:
- Don't use LDAP server

**To test**:
- Send an email in dev with ldap server set to null. Emails should appear in emails.txt if using the latest podman image.

### DIFF
--- a/api/config_sample.php
+++ b/api/config_sample.php
@@ -43,7 +43,7 @@
     # CAS CA Cert (for SSO)
     $cacert = '/etc/certs/ca-bundle.crt';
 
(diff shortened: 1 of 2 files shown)


ndg63276 (approved):
lgtm

PR #675, opened 2023-09-25
LIMS-933: Attach auth header to shipping service requests

MattPrit:
**JIRA ticket**: [LIMS-933](https://jira.diamond.ac.uk/browse/lims933)

**Summary**:

This change attaches an auth header to shipping service requests. This is either the user access token as a bearer token or basic auth using stored synchweb credentials.

**Changes**:
- Added new separate configuration parameters for shipping service API and frontend URLS.
- Added configuration parameters for shipping service basic auth credentials.
- If the user session token cookie is present, it is added as a bearer token in shipping service requests. If not present, stored basic auth credentials are used.
- AWB PDF email link has been changed to point to the frontend AWB page rather than a direct link to the PDF from the backend.

**To test**:
- Check that, when a user books a shipping service shipment through the SynchWeb web application, an access token is used in shipping service requests.
- Check that, when a shipping service shipment is booked without user SSO, basic auth is used in shipping service requests.

### DIFF
--- a/api/src/Shipment/ShippingService.php
+++ b/api/src/Shipment/ShippingService.php
@@ -10,7 +10,6 @@ class ShippingService
 {
     private $shipping_api_url;
     private $shipping_app_url;
-    private $headers;
     public const JOURNEY_TO_FACILITY = "TO_FACILITY";
     public const JOURNEY_FROM_FACILITY = "FROM_FACILITY";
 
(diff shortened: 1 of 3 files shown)


api/src/Shipment/ShippingService.php
```diff
@@ -9,18 +9,43 @@
...
+        if (isset($shipping_service_api_user) && isset($shipping_service_api_password)) {
+            $basic_auth = base64_encode($shipping_service_api_user . ":" . $shipping_service_api_password);
+            array_push($headers, "Authorization: Basic {$basic_auth}");
+            return $headers;
+        }
+
+        throw new \Exception("Shipping service auth error: no cookie found and basic auth credentials unset");
+    }
+
+    function __construct()
```
John-Holt-Tessella, line 42: Constructing this object will fail if the setting are not set and might fail if they are null. Either:
- make sure this is not constructed if the shipping service is not being used
- make sure the constructor doesn't do anything if the shipping service is not used

John-Holt-Tessella:
Do you need to update the configs for prod and staging to have the new urls in?

John-Holt-Tessella (changes requested):

John-Holt-Tessella (approved):

PR #677, opened 2023-09-27
I15_1-1318: Cannot create default shipment for visit 0

ndg63276:
**JIRA ticket**: [I15_1-1318](https://jira.diamond.ac.uk/browse/I15_1-1318)

**Summary**:

Simple sample creation uses a 'default' dewar inside a 'default' shipment. Previously these were associated with and named after visit 0, as a beamline visit may not exist. But we no longer use visit 0, so it was failing. So just use a default dewar/shipment named after the proposal.

**Changes**:
- Allow requests to api/shipment/dewars/default without a visit parameter, as long as the prop parameter is still there
- Use variables for shipment name, dewar name and barcode to allow use with either visit or prop
- Dont add anything to ShippingHasSession if not specifying visit
- Don't use visit in container name if visit is undefined

**To test**:
- Create a new simple sample in a proposal, check it appears in a shipment named eg cy35700_Shipment1
- Check dewar inside is named eg cy35700_Dewar1
- Check container inside is named eg cy35700_samples
- Create another simple sample, check it is added to same container/dewar/shipment

### DIFF
--- a/client/src/js/modules/types/xpdf/samples/views/vue-simplesample.vue
+++ b/client/src/js/modules/types/xpdf/samples/views/vue-simplesample.vue
@@ -540,7 +540,7 @@
                     })
 
                     let container = new Container({
-                        NAME: app.prop + '-' + app.visit + '_samples',
+                        NAME: app.visit ? app.prop + '-' + app.visit + '_samples' : app.prop + '_samples',
                         CAPACITY: self.capacity,
                         CONTAINERTYPE: 'Box',
                         COMMENTS: item.comments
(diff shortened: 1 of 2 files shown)


client/src/js/modules/types/xpdf/samples/views/vue-simplesample.vue
```diff
@@ -406,10 +406,8 @@
         methods: {
             async setDewarInformation() {
               // Try to retrieve the default dewar for this proposal/visit
-              // Uses the special session-0 because at this point we are not necessarily on a session
               this.defaultDewarId = await this.$store.dispatch('fetchDataFromApi', {
                 url: '/shipment/dewars/default',
```
John-Holt-Tessella, line 410: I don't understand why we don't need to send the visit or proposal here.

John-Holt-Tessella (approved):

PR #681, opened 2023-10-09
LIMS-1025: Fix lab contact search

ndg63276:
**JIRA ticket**: [LIMS-1025](https://jira.diamond.ac.uk/browse/LIMS-1025)

**Summary**:

Fix lab contact search

**Changes**:
- Create a where clause to match the passed in search term to any of the lab contact fields
- Start with a known false condition as each field adds an "or" clause

**To test**:
- Go to a list of lab contacts on a proposal and search for each field in turn
- View a single lab contact, test it displays correctly

### DIFF
--- a/api/src/Page/Contact.php
+++ b/api/src/Page/Contact.php
@@ -52,7 +52,6 @@ class Contact extends Page
                 $fields = array('pe.givenname', 'pe.familyname', 'c.cardname',
                     'l.name', 'l.address', 'l.city', 'l.country', 'pe.phonenumber', 'l.postcode');
                 $where = $where . DatabaseQueryBuilder::getWhereSearch($this->arg('s'), $fields, $args);
-                error_log($where);
             }
 
             $tot = $this->db->pq("SELECT count(c.labcontactid) as tot FROM labcontact c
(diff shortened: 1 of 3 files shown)


api/src/Page/Contact.php
```diff
@@ -46,9 +46,24 @@ function _get_contacts() {
                 $where .= ' AND c.labcontactid=:'.(sizeof($args)+1);
                 array_push($args, $this->arg('cid'));
             }
-            
 
-            $tot = $this->db->pq("SELECT count(c.labcontactid) as tot FROM labcontact c  $where", $args);
+            if ($this->has_arg('s')) {
+                $st = sizeof($args) + 1;
```
John-Holt-Tessella, line 51: This pattern is repeated everywhere, lets not do it again. Instead we should pass `s` and a list of field names to a `whereSearch` method of `WhereClauseQueryBuilder` which is probably part of, or at least used by, `DatabaseQueryBuilder`. The query builder can be used directly here or even extended to do the whole paginate search.
Would love to see the pageing code but in there too.
This function should be something like:

```
query.where(proposalId, $this->proposalid)
if cid:
  query.where(proposalId, $this->proposalid)
query.search($fields, this->arg('s')

$where = query.where_clause()
```

John-Holt-Tessella (approved):
But then it is my change :-)

ndg63276:
Looks good to me.

PR #683, opened 2023-10-20
LIMS-1068: Move discovery away from constructor, get cookie expiry dynamically

gfrn:
**JIRA ticket**: [LIMS-1068](https://jira.diamond.ac.uk/browse/LIMS-1068)

**Summary**:

- Reduce number of calls to OIDC provider by only performing discovery when necessary
- Get expiration time for token from provider, apply it to cookie

**Changes**:
- Cookie expiry is applied based on value returned from OIDC provider
- Endpoint discovery is only performed when a function is called and there are no known endpoints

**To test**:
- Check if user logout behaves as expected
- Check if user login behaves as expected
- Check if opening page in a new tab doesn't trigger login
- Check if moving between pages works

### DIFF
--- a/api/src/Authentication/Type/OIDC.php
+++ b/api/src/Authentication/Type/OIDC.php
@@ -11,7 +11,7 @@ class OIDC extends AuthenticationParent implements AuthenticationInterface
     private $providerConfig = array();
 
     private function getEndpoints() {
-        if (empty($providerConfig)) {
+        if (empty($this->providerConfig)) {
             global $sso_url, $oidc_client_id, $oidc_client_secret;
 
             $ch = curl_init();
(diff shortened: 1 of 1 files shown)


api/src/Authentication/Type/OIDC.php
```diff
@@ -10,33 +10,36 @@ class OIDC extends AuthenticationParent implements AuthenticationInterface
...
-        $newProviderConfig = json_decode($response);
-
-        if(!$newProviderConfig
-            || !isset($newProviderConfig->userinfo_endpoint)
-            || !isset($newProviderConfig->authorization_endpoint) 
-            || !isset($newProviderConfig->token_endpoint)) {
-            error_log("OIDC Authentication provider replied with invalid JSON body");
-            return;
+    private function getEndpoints() {
+        if (empty($providerConfig)) {
```
ndg63276, line 14:
```suggestion
        if (empty($this->providerConfig)) {
```

ndg63276 (approved):
LGTM

PR #686, opened 2023-10-27
LIMS-768: Modify the values saved in database for Simple Sample Uploads

ndg63276:
**JIRA ticket**: [LIMS-768](https://jira.diamond.ac.uk/browse/LIMS-768), [LIMS-650](https://jira.diamond.ac.uk/browse/LIMS-650)

**Summary**:

I15-1 have asked for the BLSample.name and Crystal.name fields to be changed for Simple Sample uploads

**Changes**:
- Use {acronym}-sample as Crystal.name
- Use {acronym} as BLSample.name

**To test**:
- Go to a 'Phase' eg /phases/pid/580719
- Click 'New Simple Sample'
- Fill Acronym as something memorable eg 'foo', and Packing Fraction as 0
- Verify new sample is created with name 'foo-sample'
- Verify new instance of sample is created with name 'foo'

### DIFF
--- a/api/src/Page/Sample.php
+++ b/api/src/Page/Sample.php
@@ -355,7 +355,7 @@ class Sample extends Page
                 $isCapillary = sizeof($crystals) > 1 ? true : false;
 
                 foreach ($crystals as $sample) {
-                    $c = array('NAME' => $protein->ACRONYM . '-sample');
+                    $c = array('NAME' => $phase->ACRONYM . '-sample');
                     foreach (array('SPACEGROUP', 'COMMENTS') as $f)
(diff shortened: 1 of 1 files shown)


api/src/Page/Sample.php
```diff
@@ -355,8 +355,8 @@ function _add_simple_sample()
                 $isCapillary = sizeof($crystals) > 1 ? true : false;
```
gfrn, line 356: Since the last element in the `foreach` is persisted outside the scope of the loop due to how PHP scoping works, it would be clearer to explicitly set `$protein`, so that it's clearer for developers in the future

```suggestion

                $protein = $phases[0];
                
```

gfrn (approved):

PR #693, opened 2023-11-21
[LIMS-1120] Display message when user is not in database

gfrn:
**JIRA ticket**: [LIMS-1120](https://jira.diamond.ac.uk/browse/LIMS-1120)

**Summary**:

Display a message in the login page when an user logs in successfully, but their username matches no entries in the database.

**Changes**:
- Check if user exists when exchanging code for token in backend
- Add message in frontend when backend returns 403

**To test**:
- Login with a valid FedID
- If possible, log in with an email address, and if not, mock out the response of `authenticateByCode` to return an invalid username.


ndg63276:
Haven't been able to test but the code looks ok to me.

PR #702, opened 2023-12-05
LIMS-1163: Improve error message when token is invalid

ndg63276:
**JIRA ticket**: [LIMS-1163](https://jira.diamond.ac.uk/browse/LIMS-1163)

**Summary**:

Users complain of not being allowed to print an air waybill after booking a shipment into Diamond.

**Changes**:
- Specify in the error message generated whether this is because the one time token is too old, or non-existent
- Add user agent and timestamp to the error message generated
- Allow increasing of 10 second age limit with an optional config variable

**To test**:
- Generate an AWB for a shipment, check can view it from the shipment page
- Set a $max_token_age = -1 config variable and check the error generated (delete it afterwards)
- View a visit pdf report and CSV download from /visits page, as these also use the sign function

### DIFF
--- a/api/src/Controllers/AuthenticationController.php
+++ b/api/src/Controllers/AuthenticationController.php
@@ -168,10 +168,11 @@ class AuthenticationController
                 $token = $token[0];
                 if ($token['AGE'] > $max_token_age)
                 {
-                    $err = 'Authorisation token too old. Age: '.$token['AGE'].'s. Max age: '.$max_token_age.'s.';
-                    $err .= ' Please press back and then try again.';
+                    $err = 'Authorisation token too old. Please press back and then try again.';
                     $err .= ' If this problem persists, please try clearing your cookies or using a different browser.';
-                    $this->returnError(400, $err, true);
+                    error_log('Authorisation token too old. Age: '.$token['AGE'].'s. Max age: '.$max_token_age.'s.');
+                    error_log('User-agent: ' . $_SERVER['HTTP_USER_AGENT']);
+                    $this->returnError(400, $err);
                 }
                 $qs = $_SERVER['QUERY_STRING'] ? (preg_replace('/(&amp;)?token=\w+/', '', str_replace('&', '&amp;', $_SERVER['QUERY_STRING']))) : null;
                 if ($qs)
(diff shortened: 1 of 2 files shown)


api/src/Controllers/AuthenticationController.php
```diff
@@ -300,7 +317,14 @@ private function returnResponse($code, $message)
 
     private function returnError($code, $message)
     {
-        $this->returnResponse($code, array('error' => $message));
+        $this->returnResponse(
+            $code,
+            array(
+                'error' => $message,
+                'user-agent' => $_SERVER['HTTP_USER_AGENT'],
+                'timestamp' => gmdate('c', time())
```
MattPrit, line 325: I don't think we need to return these to the user. I think we should instead just log these when we log the auth token failure error and reason.

api/src/Controllers/AuthenticationController.php
```diff
@@ -178,13 +187,21 @@ private function processOneTimeUseTokens(): bool
                     $need_auth = false;
                     $this->dataLayer->deleteOneTimeUseToken($tokenId);
                 }
+                else
+                {
+                    $err = 'Authorisation token not valid for this URL.';
```
MattPrit, line 192: I'm not sure that we would want to tell the user this. Perhaps we should, if a valid token is not found, return a more generic message to the user, e.g. "Invalid one-time authorization token, please go back and try again...", but log the reason internally.

api/src/Controllers/AuthenticationController.php
```diff
@@ -157,16 +157,25 @@ private function checkAuthRequiredForSpecificSituations($parts): bool
 
     private function processOneTimeUseTokens(): bool
     {
+        global $max_token_age;
```
MattPrit, line 160: I'm not convinced that there is a need to make this configurable. Perhaps we should wait until error logs suggest that the max token age is responsible for errors?

MattPrit (changes requested):

api/src/Controllers/AuthenticationController.php
```diff
@@ -161,12 +161,18 @@ private function processOneTimeUseTokens(): bool
...
-            # Remove tokens more than 10 seconds old, they should have been used
-            $this->dataLayer->deleteOldOneTimeUseTokens();
+            $max_token_age = 10;
             $token = $this->dataLayer->getOneTimeUseToken($tokenId);
             if (sizeof($token))
             {
                 $token = $token[0];
+                if ($token['AGE'] > $max_token_age)
+                {
+                    $err = 'Authorisation token too old. Age: '.$token['AGE'].'s. Max age: '.$max_token_age.'s.';
```
gfrn, line 171: I don't think this information should be returned to the user

PR #704, opened 2023-12-11
LIMS-1136: Trim preceding spaces from postcode and email address

ndg63276:
**JIRA ticket**: [LIMS-1136](https://jira.diamond.ac.uk/browse/LIMS-1136)

**Summary**:

Currently, it is possible for post codes to contain spaces before and after the actual post code. Sending such post codes to the shipping service results in a 422 error.

**Changes**:
- Use `trim` rather than `rtrim`, for post code and email address.

**To test**:
- Book an incoming shipment
- Book an incoming shipment with a space at the start of the post code
- Book an incoming shipment with a space at the start of the email address


MattPrit (approved):
Looks good to me

PR #707, opened 2024-01-03
LIMS-980: Add all samples to queue should obey filters

ndg63276:
**JIRA ticket**: [LIMS-980](https://jira.diamond.ac.uk/browse/LIMS-980)

**Summary**:

LIMS-219 (https://github.com/DiamondLightSource/SynchWeb/pull/591) added a button to add all subsamples to the queue, but it ignored the current filter, which can be auto/manual/point/region.

**Changes**:
- Get the id of the currently selected filter element and add it to the backend request
- Use the filter supplied to alter the where clause of the query

**To test**:
- Go to a VMXi proposal (eg nt37104), go to Containers, pick one. On the next page mark a point and a region using the "Mark Point" and "Mark Region" buttons. Click "Prepare for Data Collection"
- Filter the available samples using the Point/Region/Auto/Manual buttons, click the Add All to Queue button. Check the right number appear in the queue in the lower half of the page.
- Turn off any filters and Add All again, check every (not complete) sample is added to the queue.

### DIFF
--- a/client/src/js/modules/imaging/views/queuecontainer.js
+++ b/client/src/js/modules/imaging/views/queuecontainer.js
@@ -567,12 +567,18 @@ define(['marionette',
         queueAllSamples: function(e) {
             e.preventDefault()
 
+            var data = {}
+            current = this.$el.find('.afilt').find('.current')
+            if (current.length > 0) {
+                data.filter = current.attr('id')
+            }
+
             var self = this
             this.$el.addClass('loading');
             Backbone.ajax({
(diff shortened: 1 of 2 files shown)


client/src/js/modules/imaging/views/queuecontainer.js
```diff
@@ -567,12 +567,18 @@ define(['marionette',
         queueAllSamples: function(e) {
             e.preventDefault()
 
+            var data = {}
```
gfrn, line 570:
```suggestion
            const data = {}
```

gfrn (approved):
I think this depends on LIMS-979, but looks good to me!

Issue #53, opened 2019-08-14
Assumes one-to-one mapping of beamline to PV prefix


jlmuir:
In `api/config_sample.php`, there is a variable `$bl_pv_map` that assumes a one-to-one mapping between a beamline and a PV prefix:

```
# Map of beamlinename to pv prefix
$bl_pv_map = array(
  'i02' => 'BL02I',
  'i03' => 'BL03I',
);
```

At some sites (e.g., IMCA-CAT), there is not a one-to-one between a beamline and a PV prefix, so it would be better to just list PVs explicitly for a given beamline rather than trying to build them dynamically with a PV prefix based on the beamline.

ndg63276:
Raised as an internal issue at https://jira.diamond.ac.uk/browse/LIMS-697

jlmuir:
I didn't investigate how it's used; I was just editing the config to be appropriate for the IMCA-CAT beamline instance, and when I encountered this `$bl_pv_map` variable, I knew that I couldn't possibly edit its value to be correct for IMCA-CAT because it makes the assumption that a beamline has exactly one PV prefix, which is not the case at IMCA-CAT.

ndg63276:
Thanks @jlmuir, good to know it isn't actively breaking things, even if it seems badly written. The same is actually true at Diamond, i03 has PVs starting FE03I (for front end) and also BL03I (for optics and endstation).

We don't really want to list the PVs explicitly, as there are 37 per beamline, all of the format BLxxI-MO-ROBOT-01:PUCK_yy_NAME, where yy is between 01 and 37. The PV is to say which puck is loaded in which position, eg: 

```
BL03I-MO-ROBOT-01:PUCK_01_NAME CPS-0001
```

Do you have anything equivalent? How are your PV prefixes split, could we just give another key eg

```
$bl_pv_map = array(
    'i03' => array(
        'frontend' => 'FE03I',
        'optics' => 'BL03I',
        'endstation' => 'BL03I',
    )
);
```

jlmuir:
[...]

PR #712, opened 2024-01-10
LIMS-697: Use more specific PVs for puck names

ndg63276:
**Github issue**: https://github.com/DiamondLightSource/SynchWeb/issues/53
**JIRA ticket**: [LIMS-697](https://jira.diamond.ac.uk/browse/LIMS-697)

**Summary**:

@jlmuir points out that not all facilities have a one-to-one mapping of beamlines to pv prefixes (DLS included). As the `$bl_pv_map` variable is only used in one place, we should replace it with a template for the specific PVs it is used for.

**Changes**:
- Remove `$bl_pv_map` variable
- Remove DLS-specific hardcoded PV names
- Add `$bl_puck_names` variable for a template for the puck names PV

**To test**:
- Go to the assign container page for a visit (eg /assign/visit/cm37235-1) and check the loaded pucks are displayed next to the puck number (eg CPS-0001 below)
image

### DIFF
--- a/api/config_sample.php
+++ b/api/config_sample.php
@@ -374,10 +374,9 @@
                                            ),
     );
 
-    # Map of beamlinename to pv prefix
-    $bl_pv_map = array(
-        'i02' => 'BL02I',
-        'i03' => 'BL03I',
+    # Map of beamlinename to puck name pv
+    $bl_puck_names = array(
+        'i03' => "BL03I-MO-ROBOT-01:PUCK_%'02d_NAME"
     );
 
     # Dials server values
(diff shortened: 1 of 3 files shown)


gfrn (approved):
I think the way PV searches are done (build PV name array, then mass `caget`, and reindex, searching for the returned PVs in the initial array) feels a bit off, but I cannot say with absolute certainty it can be refactored without affecting other bits of code elsewhere.

Also, long shot, but EPICS exposes `EPICS_BASE` and `EPICS_HOST_ARCH`, so *if* we really wanted to, we could also build the path from these environment variables rather than add a path to `caget` in `config.php`, but that's minimal, I think

PR #714, opened 2024-01-19
LIMS-1188: Create web-based CSV importer

ndg63276:
**JIRA ticket**: [LIMS-1188](https://jira.diamond.ac.uk/browse/LIMS-1188)

**Summary**:

The current Diamond CSV importer requires SSH or NX access to Diamond, followed by a special Bash script. Authentication then relies on putting the file in a particular location within a visit.
This importer is web-based, so can be done without SSH or NX, and uses the ISPyB authorisation to check your proposal.
The original code is from https://github.com/DiamondLightSource/SynchWeb/pull/239, but with a bunch of changes to make it work for Diamond. (IMCA csv files do seem to still work, although it no longer creates proteins that don't exist.) Different sites can create different CSV 'formats'.

**Changes**:
- Define CSV profiles for Diamond and IMCA
- Add button on shipment page, and routing to go to the CSV importer
- Add CSV importer page html
- Javascript file browser and file drop location to get the CSV file
- Javascript logic to process the CSV file
- Some extra restrictions on energy and on screening methods/parameters
- Add ability to get all existing samples from a shipment
- Add ability to set sublocation and observed resolution on a sample

**To test**:
- Test creating a shipment, puck and samples without using CSV importer
- Using proposal mx23694 and csv_profile of "diamond", test uploading this file [mx23694.csv](https://github.com/DiamondLightSource/SynchWeb/files/13987706/mx23694.csv)
- Using proposal mx23694 and csv_profile of "imca", test uploading this file [csv_imca.csv](https://github.com/DiamondLightSource/SynchWeb/files/13987682/csv_imca.csv)
- Modify the mx23694.csv file so that the protein acronym is different (eg RFH3zxcvbnm), or location is not 1-16, or multiple rows have the same location, or mulitple rows have the same protein and sample name, or screening method is something invalid. Upload should be rejected without populating the form.
- Modify the mx23694.csv file so that some other field is invalid, eg unit cell using letters, upload should be accepted but fields in the form should be red and the import shouldn't be allowed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,7 @@ This file should be copied to create a `client/src/js/config.json` file and edit
 | site_name | Site Name to display in footer |
 | site_link | URL to site home page |
 | data_catalogue | Object that includes name and url property for a link to a data catalogue - displayed on the landing page |
+| csv_profile | The csv profile for importing shipments, currently diamond or imca, as client/src/js/csv/<csv_profile>.js |
 
 Site Image can be customised via the tailwind.config.js header-site-logo and footer-site-logo values.
 
--- a/api/src/Page/Sample.php
+++ b/api/src/Page/Sample.php
@@ -52,6 +52,7 @@ class Sample extends Page
         'CRYSTALID' => '\d+',
         'CONTAINERID' => '\d+',
         'LOCATION' => '\d+',
+        'SUBLOCATION' => '\d+',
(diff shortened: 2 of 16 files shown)


ndg63276:
After discussion with James and Karl, the conclusion was to wait for some new API endpoints, so converting to draft for now.

PR #720, opened 2024-01-26
LIMS-1185: Fix incorrect total dose displayed

ndg63276:
**JIRA ticket**: [LIMS-1185](https://jira.diamond.ac.uk/browse/LIMS-1185) and [LIMS-1212](https://jira.diamond.ac.uk/browse/LIMS-1212)

**Summary**:

The total dose displayed on data collection groups is incorrect, and also shown to a huge number of decimal places.

**Changes**:
- Limit the totaldose field to 2 decimal places, as done for other fields
- Use the totaldose returned from the backend, which does a sum of the total absorbed dose, rather than getting one value and multiplying by the number of data collections.
- Also limit chi/phi/overlap to 1 decimal place

**To test**:
- View a data collection group, eg /dc/visit/cm33903-5/dcg/10774174, check the dose displayed for each data collection is shown to 2 decimal places. Add together the total dose for the group (11.40MGy)
- Find the same group on the parent view (eg dc/visit/cm33903-5/ty/ page 10), check the total dose displayed is correct (11.40MGy) with 2 decimal places

### DIFF
--- a/api/src/Page/DC.php
+++ b/api/src/Page/DC.php
@@ -975,7 +975,7 @@ class DC extends Page
 
             // Data collections
             if ($dc['TYPE'] == 'data') {
-                $nf = array(1 => array('AXISSTART'), 2 => array('RESOLUTION', 'TRANSMISSION', 'AXISRANGE', 'TOTALDOSE'), 4 => array('WAVELENGTH', 'EXPOSURETIME'));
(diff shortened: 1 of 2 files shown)


gfrn:
The second test displays 12.96MGy, but I assume this is fine, since 12.96 MGy is the sum of all data collections in the data collection group in page 10 (1.5 + 1.0 + 4.0 + 6.46044). Does that check out?

ndg63276:
Yeah, looks like I messed up the link, was supposed to be /dc/visit/cm33903-5/ty/fc, page 10, 8th data collection down. It's a bit hard to link to as the actual permalink takes you inside the group, rather than showing the group summary.

Anyway, obviously it should work for any data collection group, as long as the Total Dose displayed on the main page matches the sum of the doses of each data collection, that's all that matters :-)

gfrn (approved):

PR #722, opened 2024-02-01
LIMS-904: Improve help text for internal transfer requests

ndg63276:
**JIRA ticket**: [LIMS-904](https://jira.diamond.ac.uk/browse/LIMS-904)

**Summary**:

Give users a bit more guidance if they require more data collection time.

**Changes**:
- Edited the post-visit email to say "Or fill out the dewar transfer form if you require more data collection time:" instead of "Or fill out the dewar transfer form if you would like it to be transferred to another beamline for another visit"
- Removed instruction to print this email
- Edited the transfer request form help text under 'Next Visit' to say 'Select "-" to return the dewar to the responsive scheduling queue'
- Removed the comments field from the transfer request form, as they are not recorded anywhere

**To test**:
- Fill out a dewar transfer request, check there is no comments field, check it correctly sends an email

### DIFF
--- a/api/assets/emails/html/storage-rack.html
+++ b/api/assets/emails/html/storage-rack.html
@@ -13,7 +13,7 @@
 				</ul>
 			<?php } ?>
 			<p>Please either fill out the dewar dispatch form if you would like your dewar to be returned to you:<br /><a href="https://ispyb.diamond.ac.uk/dewars/dispatch/<?php echo $data['DEWARID'] ?>">https://ispyb.diamond.ac.uk/dewars/dispatch/<?php echo $data['DEWARID'] ?></a></p>
-			<p>Or fill out the dewar transfer form if you would like it to be transferred to another beamline for another visit:<br /><a href="https://ispyb.diamond.ac.uk/dewars/transfer/<?php echo $data['DEWARID'] ?>">https://ispyb.diamond.ac.uk/dewars/transfer/<?php echo $data['DEWARID'] ?></a></p>
+			<p>Or fill out the dewar transfer form if you require more data collection time:<br /><a href="https://ispyb.diamond.ac.uk/dewars/transfer/<?php echo $data['DEWARID'] ?>">https://ispyb.diamond.ac.uk/dewars/transfer/<?php echo $data['DEWARID'] ?></a></p>
 			<p>Your dewar will be topped up with liquid nitrogen in the meantime.</p>
 		</div>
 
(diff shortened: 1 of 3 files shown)


api/assets/emails/html/storage-rack.html
```diff
@@ -13,7 +13,7 @@ <h2>Dear <?php echo $data['LCRET'] ?></h2>
 				</ul>
 			<?php } ?>
 			<p>Please either fill out the dewar dispatch form if you would like your dewar to be returned to you:<br /><a href="https://ispyb.diamond.ac.uk/dewars/dispatch/<?php echo $data['DEWARID'] ?>">https://ispyb.diamond.ac.uk/dewars/dispatch/<?php echo $data['DEWARID'] ?></a></p>
-			<p>Or fill out the dewar transfer form if you would like it to be transferred to another beamline for another visit:<br /><a href="https://ispyb.diamond.ac.uk/dewars/transfer/<?php echo $data['DEWARID'] ?>">https://ispyb.diamond.ac.uk/dewars/transfer/<?php echo $data['DEWARID'] ?></a></p>
+			<p>Or fill out the dewar transfer form if you require more data collection time:<br /><a href="https://ispyb.diamond.ac.uk/dewars/transfer/<?php echo $data['DEWARID'] ?>">https://ispyb.diamond.ac.uk/dewars/transfer/<?php echo $data['DEWARID'] ?></a></p>
```
gfrn, line 16: Should we keep parts of the old message mentioning transfers can be performed through this page? Something among the lines of "[...] if you require more data collection time, or if you would like it to be transferred to another beamline for another visit"

ndg63276: Originally a dewar transfer meant someone would physically carry the dewar to another beamline, whereas now they are all in the storage rack, and we are just allocating visits to dewars in the database. So I tried to keep the message general (even applicable to non-beamlines).

gfrn: That makes sense! Thanks for clarifying

client/src/js/templates/shipment/transfer.html
```diff
@@ -81,7 +81,7 @@ <h1>Request Internal Dewar Transfer</h1>
             <li class="head">Next Visit Details</li>
             <li>
                 <label>Next Visit
-                    <span class="small">The next visit the dewar is going to be used for</span>
+                    <span class="small">Select "-" to return the dewar to the responsive scheduling queue</span>
```
gfrn, line 84: Word this as "leave blank" instead of "select -" maybe? Or we could change the default select option to "return to responsive scheduling queue", if that doesn't break anything else elsewhere

ndg63276: I like "Leave Blank", it looks weird with the quotes around the dash. I thought changing the option text would look a little odd, and there might be other reasons to leave it blank as well.

gfrn (approved):

Issue #217, opened 2021-04-13
API GET /proposal/visits/{visit} endpoint requires prop param but not documented


jlmuir:
The GET `/proposal/visits/{visit}` endpoint requires the `prop` parameter, but the Swagger API spec doc does not list it as a parameter nor indicate that it is required.  For example, the GET request

```
/api/proposal/visits/sw12345-1
```

fails with the following exception:

```
exception 'ErrorException' with message 'Undefined index: prop' in /www/api/src/Page.php:518
Stack trace:
#0 /www/api/src/Page.php(518): Slim\Slim::handleErrors(8, 'Undefined index...', '/www/api/src/P....', 518, Array)
#1 /www/api/src/Page/Proposal.php(291): SynchWeb\Page->arg('prop')
#2 [internal function]: SynchWeb\Page\Proposal->_get_visits('sw12345-1')
#3 /www/api/vendor/slim/slim/Slim/Route.php(462): call_user_func_array(Array, Array)
#4 /www/api/vendor/slim/slim/Slim/Slim.php(1326): Slim\Route->dispatch()
#5 /www/api/vendor/slim/slim/Slim/Middleware/Flash.php(85): Slim\Slim->call()
#6 /www/api/vendor/slim/slim/Slim/Middleware/MethodOverride.php(92): Slim\Middleware\Flash->call()
#7 /www/api/vendor/slim/slim/Slim/Slim.php(1271): Slim\Middleware\MethodOverride->call()
#8 /www/api/src/Dispatch.php(47): Slim\Slim->run()
#9 /www/api/index.php(85): SynchWeb\Dispatch->dispatch()
#10 {main}
```

However, it succeeds if the `prop` parameter is specified in the GET request:

```
/api/proposal/visits/sw12345-1?prop=sw12345
```

ndg63276:
Forgive my ignorance but how do I see the Swagger spec doc?

ndg63276:
Have made an internal ticket ([LIMS-1138](https://jira.diamond.ac.uk/browse/LIMS-1138) to work on this.

PR #723, opened 2024-02-05
LIMS-1138: Update Swagger spec

ndg63276:
**JIRA ticket**: [LIMS-1138](https://jira.diamond.ac.uk/browse/LIMS-1138)

**Summary**:

The docs for the Synchweb API are outdated in a few places, this PR is to update them. As first reported in https://github.com/DiamondLightSource/SynchWeb/issues/217.

**Changes**:
- Remove required prop parameter from /proposal (get)
- Add required prop parameter to /proposals/visits/visit (get/patch) and /dc/{id} (get)
- Update endpoint for /dc/{id}, /dc/single/t/{type}/{id}, /processing/status, /processing/{id}
- Remove non-existent get on /dc/single/t/{type}/{id}
- Fix docs on post to /dc/comment
- Remove non-existent event types

**To test**:
- Go to https://editor.swagger.io/, load spec.json

### DIFF
--- a/api/docs/dist/spec.json
+++ b/api/docs/dist/spec.json
@@ -3016,7 +3016,7 @@
         }
       }
     },
-    "/dc/comments/{dcid}": {
+    "/dc/comments/{dccid}": {
       "get": {
         "security": [
           {
@@ -3029,7 +3029,7 @@
         "summary": "Get a single data collection comment",
         "parameters": [
           {
-            "name": "dcid",
+            "name": "dccid",
             "in": "path",
             "description": "Data collection comment id",
             "type": "number",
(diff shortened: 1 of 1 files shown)


api/docs/dist/spec.json
```diff
@@ -3028,7 +3016,7 @@
         }
       }
     },
-    "/dc/comments/{dccid}": {
+    "/dc/comments/{dcid}": {
```
gfrn, line 3019: Wouldn't `dccid` stand for `data collection comment ID` in this case, and therefore we should keep it? (These abbreviations/acronyms are immensely confusing...)

ndg63276: Yep, I got confused by the fact that in the Synchweb code it is called dcid (https://github.com/DiamondLightSource/SynchWeb/blob/master/api/src/Page/DC.php#L1673) but it is still a data collection comment id.

gfrn (approved):

PR #724, opened 2024-02-08
LIMS-1202: Fix UAS filter and table sorting on Phases/Proteins page

ndg63276:
**JIRA ticket**: [LIMS-1202](https://jira.diamond.ac.uk/browse/LIMS-1202)

**Summary**:

On an xpdf proposal, on the Phases page, there is a tickbox to "Show Only UAS Samples". This has the effect of grouping phases by their external id, so only one entry appears for each UAS sample, but as they are ordered by most recent, the recent non-UAS phases are shown instead of older UAS phases.
Sorting on the same table only worked for a couple of columns as well.

**Changes**:
- When the "Show only UAS Samples" tickbox is ticked, only show Phases (Proteins) with the "proteinType" of "ORIGIN:UAS"
- Add more possible columns to sort by in the database query
- Get the number of samples and number of data collections in the original query so they can be used for sorting as well
- Remove redundant queries to get dcount and scount, and convoluted way to get sequence, this already comes from the original query.

**To test**:
- Go to an xpdf proposal eg cm37261 and go to the Phases page from the main menu. Tick and untick the 'Show only UAS Samples' box, check all UAS samples have a risk rating.
- Click on a Phase, then click Create New Simple Sample, put in a new acronym. Go back to the phases page and check the new acronym is only shown when 'Show only UAS samples' is unticked.
- Check sorting by clicking on the table headers works.
- Go to an mx proposal eg mx23694, go to the Proteins page. Check sorting by clicking on the table headers works, including sample count and data collection count.

### DIFF
--- a/api/src/Page/Sample.php
+++ b/api/src/Page/Sample.php
@@ -1727,12 +1727,24 @@ class Sample extends Page
 
         // Only display original UAS approved proteins (not clones which have the same externalId)
         if ($this->has_arg('original') && $this->arg('original') == 1) {
-            $group = 'pr.externalId';
-            $order .= ', pr.bltimeStamp DESC';
+            $where .= ' AND pr.proteintype = "ORIGIN:UAS"';
         }
 
         if ($this->has_arg('sort_by')) {
-            $cols = array('NAME' => 'pr.name', 'ACRONYM' => 'pr.acronym', 'MOLECULARMASS' => 'pr.molecularmass', 'HASSEQ' => "CASE WHEN sequence IS NULL THEN 'No' ELSE 'Yes' END");
+            $cols = array(
+                'NAME' => 'pr.name',
+                'ACRONYM' => 'pr.acronym',
+                'MOLECULARMASS' => 'pr.molecularmass',
+                'DENSITY' => 'pr.density',
+                'SEQUENCE' => 'pr.sequence',
+                'SAFETYLEVEL' => 'pr.safetylevel',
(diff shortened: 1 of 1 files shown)


api/src/Page/Sample.php
```diff
@@ -1748,56 +1760,23 @@ function _proteins()
...
 
                                 FROM protein pr
                                 LEFT OUTER JOIN concentrationtype ct ON ct.concentrationtypeid = pr.concentrationtypeid
                                 LEFT OUTER JOIN componenttype cmt ON cmt.componenttypeid = pr.componenttypeid
                                 LEFT OUTER JOIN protein_has_pdb php ON php.proteinid = pr.proteinid
-                                /*LEFT OUTER JOIN crystal cr ON cr.proteinid = pr.proteinid
+                                LEFT OUTER JOIN crystal cr ON cr.proteinid = pr.proteinid
                                 LEFT OUTER JOIN blsample b ON b.crystalid = cr.crystalid
-                                LEFT OUTER JOIN datacollection dc ON b.blsampleid = dc.blsampleid*/
+                                LEFT OUTER JOIN datacollection dc ON b.blsampleid = dc.blsampleid
```
gfrn, line 1773: `BLSAMPLEID` as a column is possibly deprecated, could we perform this join in another way? Possibly through the `blSubSampleId` column, if that is not always populated

Also, could we perform said join as a series of inner joins, for performance reasons?

gfrn (approved):

PR #725, opened 2024-02-09
LIMS-303: Show visits from other villages on user calendar

ndg63276:
**JIRA ticket**: [LIMS-303](https://jira.diamond.ac.uk/browse/LIMS-303)

**Summary**:

The Calendar view currently only shows visits from the same beamline 'type' as the active proposal. Users would like to be able to see their visits from all villages. Don't apply to staff as the visit list becomes unwieldy.

Also, the visit list is requested 3 times when the page is first loaded.

**Changes**:
- Only specify the beamline type if a staff member is logged in
- Dont fetch the new visit list when the current year/month is first set, there is an explicit fetch in the `mounted()` function

**To test**:
- Log in as a staff member, check the calendar page displays as before, but there is only one request to `/api/proposal/visits`
- Log in as a user who has visits in different villages (eg ygr68196), check the calendar page now shows all visits from different beamline types (eg i04, m02)

### DIFF
--- a/client/src/js/modules/calendar/views/calendar-view.vue
+++ b/client/src/js/modules/calendar/views/calendar-view.vue
@@ -244,7 +244,7 @@ export default {
           all: 1,
         }
 
-        queryParams.ty = this.proposalType
+        if (app.staff) queryParams.ty = this.proposalType
         if (this.selectedBeamline !== 'all') queryParams.bl = this.selectedBeamline
 
         const visitsCollection = new Visits(null, {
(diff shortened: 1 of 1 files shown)


gfrn:
Although the PR itself works, I have noticed that the calendar button in the navbar does not appear for anyone that isn't staff, as there is a `isStaff` check in the [header.vue](https://github.com/DiamondLightSource/SynchWeb/blob/master/client/src/js/app/components/header.vue#L23) component. Should we remove that and make the calendar visible for everyone?

ndg63276:
Yeah, was going to do that as part of LIMS-1207. Non-stafff can get to the calendar at the moment from the menu.

gfrn (approved):

Issue #316, opened 2022-07-25
"Unknown column" DB error on sample page


ejd53:
Navigate to sample page, e.g., https://ispyb.diamond.ac.uk/samples/sid/3909536
Scroll to "Data Collections"
Click "Processing Errors"
See error at top of page:

**A database error has occurred
Unknown column 'app.datacollectionid' in 'on clause'359**

Request URI: https://ispyb.diamond.ac.uk/api/dc?page=1&per_page=15&total_pages=0&total_entries=0&sid=3909536&pp=5&t=err&prop=mx26794
Responds with 503 Service Unavailable, body {"title":"There was an error with MySQL","msg":"Unknown column 'app.datacollectionid' in 'on clause'359"}

ndg63276:
I think this was fixed when https://github.com/DiamondLightSource/SynchWeb/pull/251 was merged in last October.

PR #726, opened 2024-02-12
LIMS-1219: Fix slow data collection queries

gfrn:
**JIRA ticket**: [LIMS-1219](https://jira.diamond.ac.uk/browse/LIMS-1219)

**Summary**:

The processing errors tab in the data collections page performed an outer join that effectively evaluated to outer joining on `true`, creating an unperformant query that took up lots of resources and often timed out.

Also, the query that returns the total number of pages performed `UNION`s instead of `UNION ALL`s, which could result in an unexpectedly smaller page count. Due to deduping.

**Changes**:
- Rewrote two portions of the query

**To test**:
- Go to a data collection (/dc/visit/cm33866-1/ty/dc), check if the page loads
- Go to a data collection processing errors tab (dc/visit/cm33866-1/ty/err), check if the page times out
- Switch to the new changes, observe that now the processing errors tab should load
- Check the requests made to the `dc` endpoint in the developer console, and guarantee they match the ones in the previous version

### DIFF
--- a/api/src/Page/DC.php
+++ b/api/src/Page/DC.php
@@ -151,7 +151,7 @@ class DC extends Page
                 $where = " AND appm.autoprocprogrammessageid IS NOT NULL AND (appm.severity = 'WARNING' OR appm.severity = 'ERROR')";
                 $extj[0] .= "LEFT OUTER JOIN autoprocintegration api ON dc.datacollectionid = api.datacollectionid
                         LEFT OUTER JOIN processingjob pj ON dc.datacollectionid = pj.datacollectionid
-                        LEFT OUTER JOIN autoprocprogram app ON (app.autoprocprogramid = api.autoprocprogramid OR dc.datacollectionid = pj.datacollectionid)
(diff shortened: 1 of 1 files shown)


ndg63276:
The error seems to have been introduced in https://github.com/DiamondLightSource/SynchWeb/pull/251. Also worth checking if this works on the 'Processing errors' tab on a sample page (eg https://ispyb.diamond.ac.uk/samples/sid/3909536)...

gfrn:
> The error seems to have been introduced in #251. Also worth checking if this works on the 'Processing errors' tab on a sample page (eg https://ispyb.diamond.ac.uk/samples/sid/3909536)...

Just tested it with https://localtest.diamond.ac.uk/samples/sid/4301801 and it works, also works for the one you linked (although it displays an empty list as it only has one robot load and one data collection)

ndg63276:
Thanks for checking, I used the sample URL from https://github.com/DiamondLightSource/SynchWeb/issues/316 without seeing if it was a suitable example.

PR #730, opened 2024-02-15
Handle -ve rotations, null overlap in sm/DC

ndevenish:
**JIRA ticket**: [LIMS-1232](https://jira.diamond.ac.uk/browse/LIMS-1232)

**Summary**:

I19 is now doing small molecule serial, where every other collection is run with backwards oscillation. Unfortunately, recording this accurately in ISPyB doesn't show the autoprocessing sections in Synchweb.

This changes the check to be truthy (nonzero) instead of purely positive.

Also fixes an adjacent issue where a null value for overlap was passing the intended non-zero check (null != 0). This is made truthy also.

**Changes**:
- Altered data checks to check truthiness instead of comparisons to zero

**To test**:
- Viewing https://ispyb.diamond.ac.uk/dc/visit/cy35300-1/dcg/11242587 should show "Auto Processing" for both collections, instead of the current situation:
image

### DIFF
--- a/client/src/js/templates/types/sm/dc/dc.html
+++ b/client/src/js/templates/types/sm/dc/dc.html
@@ -43,8 +43,8 @@
     </ul>
     
     <div class="holder">
-    <% if (AXISRANGE > 0) { %>
-        <% if (OVERLAP != 0) { %>
+    <% if (AXISRANGE) { %>
+        <% if (OVERLAP) { %>
         <h1 title="Click to show EDNA/mosflm/xia2 strategies" class="strat">Strategies<span><i class="fa fa-spinner fa-spin"></i></span></h1>
         <div class="strategies"></div>
         <% } else { %>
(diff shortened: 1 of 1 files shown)


client/src/js/templates/types/sm/dc/dc.html
```diff
@@ -43,8 +43,8 @@ <h1 class="title"></h1>
     </ul>
     
     <div class="holder">
-    <% if (AXISRANGE > 0) { %>
-        <% if (OVERLAP != 0) { %>
+    <% if (AXISRANGE) { %>
```
ndg63276, line 46:
```suggestion
    <% if (parseFloat(AXISRANGE)) { %>
```

client/src/js/templates/types/sm/dc/dc.html
```diff
@@ -43,8 +43,8 @@ <h1 class="title"></h1>
...
+        <% if (OVERLAP) { %>
```
ndg63276, line 47:
```suggestion
        <% if (parseFloat(OVERLAP)) { %>
```

ndg63276 (changes requested):
The values appear to be strings, so even zeroes are truthy.

ndevenish:
This appears to work fine!
image

ndg63276 (approved):

PR #698, opened 2023-11-28
LIMS-1113: Shipping service dispatch redirect

MattPrit:
**JIRA ticket**: [LIMS-1113](https://jira.diamond.ac.uk/browse/lims-1113)

**Summary**:
Adds the option to redirect to the shipping service to book shipments when requesting dispatch of appropriate Dewars.

**Changes**:
- Adds feature flag `$use_shipping_service_redirect` which enables/disables this feature
- On requesting Dewar dispatch, create a shipment request in the shipping service and redirect to the appropriate shipment workflow to book a shipment
- Use minimised dispatch request email for shipments booked via the new pathway

**To test**:
- Check that the appropriate dispatch form is displayed for academic, industrial, international users
- Check that the appropriate shipment request is created in the shipping service when requesting dispatch of a domestic academic Dewar
- Check that the minified dispatch-lite email template is used for such dispatch requests; ensure this has an appropriate link to where a Goods Handling operative can view and print the air waybill
- Check that the dispatch button links to the shipping service once a dispatch shipment request has been successfully created.

PR #731, opened 2024-02-20
ULIMS-50: Shipping service incoming shipment redirect

MattPrit:
**JIRA ticket**: [ULIMS-50](https://jira.diamond.ac.uk/browse/ulims-50)

Depends on #698 

**Summary**:

Adds the option to redirect to the shipping service to book shipments when booking an incoming shipment.

**Changes**:
- Adds feature flag `$use_shipping_service_redirect_incoming_shipments` which enables/disables this feature
- Reduce the number of fields in the create AWB form in the case of a domestic shipment when terms have been accepted
- On requesting a booking for an incoming shipment, create a shipment request in the shipping service and redirect to the appropriate shipment workflow to book a shipment
- If a shipment request has been created for the incoming shipment, display a 'manage shipment booking' button that links to the shipment request in the shipping service rather than to the 'Print Air Waybill' button.

**To test**:
- To display the new shipment view, go to an active shipment, accept the terms and conditions stipulated for shipping with Diamond's courier
- Check that the appropriate 'create awb' form is displayed for academic, domestic, industrial, international shipments
- Check that the appropriate shipment request is created in the shipping service when a user submits the form
- Check that the 'manage shipment booking' button links to the shipping service correctly once a shipment request has been successfully created

Note:
- Requires the addition of the `externalShippingIdToSynchrotron` column to the `Shipping` table in ispyb. Since the `_get_shipments` endpoint now attempts to return this column, a 500 will result whenever this endpoint is called without the new column having been added
- The changes contained in this PR (and #698) only result in shipping service redirects for domestic academic shipments that have accepted the shipping terms. All other kinds of shipment booking actions are still performed in SynchWeb. Since knowledge of a user's country is known ahead of the redirect, there is still a dependency on lab contacts containing addresses.


gfrn:
Just so we have it in writing, what is the purpose of `externalShippingIdToSynchrotron` and why should it exist in both `Dewar` and `Shipping`?

MattPrit:
> Just so we have it in writing, what is the purpose of `externalShippingIdToSynchrotron` and why should it exist in both `Dewar` and `Shipping`?

`externalShippingIdToSynchrotron` exists only in the `Shipping` table, it holds the id of the entity in the external shipping service that corresponds to the booking of  `Shipping`'s Dewars to the synchrotron. `externalShippingIdFromSynchrotron` exists only in the `Dewar` table, it holds the id of the entity in the external shipping service that corresponds to the booking of each  `Dewar` from the synchrotron. These columns are split across multiple tables because, while all Dewars in a shipment are shipped to the synchrotron together, they are dispatched individually. Alternatively, the `externalShippingIdToSynchrotron` column could be in the Dewar table, but the same id would be duplicated for each Dewar in the 'shipping'.

The required database change has now been made, see https://github.com/DiamondLightSource/ispyb-database/pull/228

client/src/js/modules/shipment/views/createawb.js
```diff
@@ -134,7 +138,12 @@ define(['backbone',
         _lcFields: ['PHONENUMBER', 'EMAILADDRESS', 'LABNAME', 'ADDRESS', 'CITY', 'POSTCODE','COUNTRY', 'GIVENNAME', 'FAMILYNAME'],
 
         toggleFacilityCourier: function() {
-            if (app.options.get('facility_courier_countries').indexOf(this.lc.get('COUNTRY')) > -1 || app.options.get('facility_courier_countries_nde').indexOf(this.lc.get('COUNTRY')) > -1 ) {
+            const shipmentCountry = this.ui.shipmentCountry.val();
+            console.log("SHIPMENT COUNTRY", this.ui.shipmentCountry);
```
gfrn, line 142: Is this `console.log` here deliberately?

client/src/js/modules/shipment/views/createawb.js
```diff
@@ -184,16 +202,14 @@ define(['backbone',
             this.quotes.queryParams.sid = this.shipment.get('SHIPPINGID')
 
             this.shipment.validation = JSON.parse(JSON.stringify(this.shipment.__proto__.validation))
+            console.log("VALIDATION", JSON.stringify(this.shipment.__proto__.validation))
```
gfrn, line 205: Is this one supposed to be here?

api/src/Page/Shipment.php
```diff
@@ -2761,9 +2933,32 @@ function _create_awb()
...
+            try {
+                $this->_create_shipment_shipment_request($ship, $dewars);
+                $this->_output(array('EXTERNAL' => "1"));
+                return;
+            } catch (\Exception $e) {
+                $error_response = json_decode($e->getMessage());
+                $this->_error($error_response->content->detail, $error_response->status);
+            }
+        }
+        $this->_error(json_encode($dewars));
```
gfrn, line 2961: Should this error message contain more data? Right now it only contains a list of the dewars sent

client/src/js/modules/shipment/views/createawb.js
gfrn, line 142:
```suggestion
```

client/src/js/modules/shipment/views/createawb.js
gfrn, line 205:
```suggestion
```

api/src/Page/Shipment.php
```diff
@@ -2761,9 +2933,32 @@ function _create_awb()
...
+            && in_array($this->arg('COUNTRY'), $facility_courier_countries)
+            && Utils::getValueOrDefault($use_shipping_service_redirect_incoming_shipments)
+        ) {
+            try {
+                $this->_create_shipment_shipment_request($ship, $dewars);
+                $this->_output(array('EXTERNAL' => "1"));
+                return;
+            } catch (\Exception $e) {
+                $error_response = json_decode($e->getMessage());
+                $this->_error($error_response->content->detail, $error_response->status);
```
gfrn, line 2958:
```suggestion
                // TODO: Use null access operator when we upgrade to PHP 8
                $error_response = json_decode($e->getMessage())->content->detail ?? $e->getMessage();
                $this->_error($error_response, $error_response->status ?? 400);
```

gfrn:
There is a possible non-issue, where if you manually delete the package description/shipping date, they cause validation to fail as they are still required, unlike all the other fields.

Not sure if this is an issue, since this would require users to deliberately "break" their own shipment, and I don't see a way this could happen during the "automated" process

gfrn (approved):

PR #739, opened 2024-03-11
LIMS-1255: Fix problematic query on visit stats page

ndg63276:
**JIRA ticket**: [LIMS-1255](https://jira.diamond.ac.uk/browse/LIMS-1255)

**Summary**:

It was noticed the visit stats page doesn't load for some visits eg https://ispyb.diamond.ac.uk/stats/visit/mx30106-12. This is caused by the CTF query doing unnecessary joins to Proposal and v_run, which are only referenced in $where if a visit is not supplied.
On further inspection, the whole CTF query is not necessary, this is only for EM which is now using PaTo. So I have removed the query and the JS that references the output.

**Changes**:
- Remove unnecessary joins to Proposal and v_run from some queries
- Remove entire CTF query and data manipulation
- Remove references to the 'scatters' data

**To test**:
- Load visit stats page for an mx visit, check everything appears correctly

### DIFF
--- a/api/src/Page/Vstat.php
+++ b/api/src/Page/Vstat.php
@@ -167,16 +167,16 @@ class Vstat extends Page
             $sched = array();
 
             $newargs = array($info['SID'], $info['SID']);
-            $missed = $this->db->pq("SELECT COUNT(smp.name) as samplecount,
-                GROUP_CONCAT(smp.name SEPARATOR ', ') as missed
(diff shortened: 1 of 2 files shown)


gfrn (approved):
Still feels a bit slow in commissioning proposals, but it's a huge leap forward regardless!

gfrn (approved):

PR #743, opened 2024-03-14
LIMS-1206: Remove stacked usage bar chart from visit page

ndg63276:
**JIRA ticket**: [LIMS-1206](https://jira.diamond.ac.uk/browse/LIMS-1206)

**Summary**:

At the top of each visit page, there is a coloured bar showing how time was used. The same info is shown on the Visit Stats page as a pie chart. Showing it on the visit page just makes the page look busier and is an extra call to the API. 

**Changes**:
- Remove the stacked bar chart

**To test**:
- Go to a visit page, check the stacked bar chart is not there
- Check the rest of the data collections are shown as usual


gfrn:
I don't think we can do it in this commit because this branch was forked from an older commit in the prerelease branch, but the only mention of `/modules/stats/views/stack` remaining is `usage-stack.vue` in one of the EM modules. 

With the removal of EM, now I believe this file (`/modules/stats/views/stack`) can be safely deleted as well

gfrn (approved):

ndg63276:
Good idea, I have merged the newer prerelease in and deleted the file, I think that's ok?

gfrn (approved):

PR #744, opened 2024-03-14
LIMS-1261: Fix the grid scan heat maps on the VMXi container view

ndg63276:
**JIRA ticket**: [LIMS-1261](https://jira.diamond.ac.uk/browse/LIMS-1261)

**Summary**:

On the container view, grid scan heatmaps have been broken by linking to GridInfo by dataCollectionId rather than dataCollectionGroupId.

**Changes**:
- Add another join to GridInfo using dataCollectionId, then use those values if the original ones are null

**To test**:
- Find a 2024 container with grid scans from VMXi (eg /containers/cid/290064/iid/125013/ well 19), the grid area should be shown on the crystal image, with a heatmap of diffraction over it.
- Check the same for a 2023 container (eg /containers/cid/280354/iid/117934 well 56)

### DIFF
--- a/api/src/Page/Download.php
+++ b/api/src/Page/Download.php
@@ -355,13 +355,15 @@ class Download extends Page
             array_push($args, $this->arg('blsampleid'));
         }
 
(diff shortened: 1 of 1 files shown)


gfrn:
A shame we're doing so much with this query we can't just group the columns by ID...

Still, unrelated to this fix, I just realised it's possible to get the wrong well to show by rapidly switching between wells that haven't been loaded yet, but I don't think this is a massive issue

gfrn (approved):

PR #752, opened 2024-04-08
LIMS-1246: Allow users with all_dewars permission to view all shipments

ndg63276:
**JIRA ticket**: [LIMS-1246](https://jira.diamond.ac.uk/browse/LIMS-1246)

**Summary**:

Members of goods handling have the "all_dewars" permission in ISPyB. We should allow them to see all shipments, and print the shipping labels. They should also see the "Logistics" page (/dewars/overview) with the truck icon in the top right.

Unfortunately this PR also allows them to see:
- A list of all proposals (could be removed but seems useful so they can navigate to a specific shipment)
- A list of all samples for a proposal
- A list of all proteins for a proposal
- A list of all lab contacts for a proposal
- Registered dewars and containers for a proposal

Does not allow them to see:
- Any visits or data collections
- Any containers
- Any details on samples, proteins, lab contacts

**Changes**:
- Set $auth to be true if the user has all_dewars permission, and the page has a prop= property and no id= or visit= properties
- Show all proposals if user has all_dewars permission
- Allow lookup of proposal for Shipping page if user has all_dewars permission
- Remove the isStaff filter on the admin menu in the top right, replace with individual permissions for each icon. Most villages only have fault reports so can be set to fault_view. 

**To test**:
- Log in as someone from goods handling, test you can navigate via a proposal to a shipment and print shipping labels
- Test a direct link to a shipment as someone from goods handling
- Test you cannot see visits or data collections
- Log in as a 'normal' user, test you can only get to shipments from your proposal
- Test a direct link to a shipment you should not be able to see
- Log in as staff, test you can see everything as normal

### DIFF
--- a/api/src/Page.php
+++ b/api/src/Page.php
@@ -366,6 +366,9 @@ class Page
                     {
                         if (sizeof(array_intersect($vis, $this->visits)))
                             $auth = True;
+                        if ($this->user->hasPermission('all_shipments'))
+                            // allow to see anything without an id or a visit
+                            $auth = True;
                     }
 
                     // No id or visit, anyone ok to view
--- a/api/src/Page.php
+++ b/api/src/Page.php
@@ -366,7 +366,7 @@ class Page
                     {
                         if (sizeof(array_intersect($vis, $this->visits)))
                             $auth = True;
-                        if ($this->user->hasPermission('all_shipments'))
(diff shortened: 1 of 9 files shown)


NKatti2011:
Not that performant on local, but works fine when deployed.

PR #756, opened 2024-04-12
LIMS-1256: Remove validation from Local Contact field

ndg63276:
**JIRA ticket**: [LIMS-1256](https://jira.diamond.ac.uk/browse/LIMS-1256)

**Summary**:

A user couldn't fill in the dispatch request form because the local contact had an "ö" in their name and the form rejected the "ö". As the user cannot edit this field any more, we should remove the validation.

**Changes**:
- Remove validation from local contact field
- Remove wwsddash pattern as it is not used anywhere else

**To test**:
- Do a dispatch request, check the form still submits ok

### DIFF
--- a/client/src/js/modules/shipment/models/dispatch.js
+++ b/client/src/js/modules/shipment/models/dispatch.js
@@ -20,13 +20,6 @@ define(['backbone'], function(Backbone) {
             pattern: 'visit',
         },
 
-        LOCALCONTACT: {
-            required: false,
-            pattern: 'wwsddash'
-        },
-
-
-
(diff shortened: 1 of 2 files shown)


gfrn:
Should we also apply this to the "full name" field in the contact details form?

ndg63276:
Potentially, but I think not right now. Obviously people might be unhappy if they cannot fill in their name with accents, umlauts etc, but it is at least editable so they can sanitise their name to fit the requirements. And we probably need some validation on the field.

gfrn (approved):

PR #759, opened 2024-04-19
LIMS-44: Speed up visit query

ndg63276:
**JIRA ticket**: [LIMS-44](https://jira.diamond.ac.uk/browse/LIMS-44)

**Summary**:

The lookup of sessionid from proposal-visit is written sub-optimally, so it has to perform a full table scan on BLSession. This changes speeds up the block of code from 40ms to 1ms.

**Changes**:
- Use regex to split the visit string into proposal code, proposal number and visit number
- Query the db with those separately

**To test**:
- Go to any visit page (eg /dc/visit/mx23694-112) and check it loads correctly, showing data collections from that visit and only from that visit
- Go to the View All Data page (/dc) and check it loads correctly

### DIFF
--- a/api/src/Page/DC.php
+++ b/api/src/Page/DC.php
@@ -177,7 +177,12 @@ class DC extends Page
         $info = array();
         # Visits
         if ($this->has_arg('visit')) {
-            $info = $this->db->pq("SELECT TO_CHAR(s.startdate, 'HH24') as sh, TO_CHAR(s.startdate, 'DDMMYYYY') as dmy, s.sessionid, s.beamlinename as bl FROM blsession s INNER JOIN proposal p ON (p.proposalid = s.proposalid) WHERE CONCAT(p.proposalcode, p.proposalnumber, '-', s.visit_number) LIKE :1", array($this->arg('visit')));
+            $pattern = '/([A-z]+)(\d+)-(\d+)/';
+            preg_match($pattern, $this->arg('visit'), $matches);
+            if (!sizeof($matches))
+                $this->_error('No such visit');
+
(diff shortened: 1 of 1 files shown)


api/src/Page/DC.php
```diff
@@ -177,7 +177,12 @@ function _data_collections($single = null)
         $info = array();
         # Visits
         if ($this->has_arg('visit')) {
-            $info = $this->db->pq("SELECT TO_CHAR(s.startdate, 'HH24') as sh, TO_CHAR(s.startdate, 'DDMMYYYY') as dmy, s.sessionid, s.beamlinename as bl FROM blsession s INNER JOIN proposal p ON (p.proposalid = s.proposalid) WHERE CONCAT(p.proposalcode, p.proposalnumber, '-', s.visit_number) LIKE :1", array($this->arg('visit')));
+            $pattern = '/([A-z]+)(\d+)-(\d+)/';
+            preg_match($pattern, $this->arg('visit'), $matches);
+            if (!sizeof($matches))
```
gfrn, line 182: This is caught and displayed as "no such proposal", but shouldn't this check that the number of matches is equal to three, rather than greater than 0? 

Right now you could type in "mx12345" and it would be a success

gfrn (approved):

api/src/Page/DC.php
ndg63276: No, I think the regex only matches if all the regex matches, and then it is portioned into groups afterwards. So the size is always 0 or 4 ($matches includes the whole match as well as the groups)
```
[ndg63276@i03-ws001 ~]$ php -a
Interactive shell
php > function get_matches($visit) {
php {     $pattern = '/([A-z]+)(\d+)-(\d+)/';
php {     preg_match($pattern, $visit, $matches);
php {     echo(sizeof($matches));
php { }
php > 
php > get_matches('mx12345-1');
4
php > get_matches('mx12345');
0

```

gfrn: Oh yeah, I noticed that what I said made no sense and resolved the conversation because of it, but now I wonder if this is a redundant check, since `$arg_list` checks for it anyway... It isn't hurting anyone, so I guess it doesn't matter

PR #762, opened 2024-04-25
LIMS-1302 - add visit to container even if UAS fails

ndg63276:
**JIRA ticket**: [LIMS-1302](https://jira.diamond.ac.uk/browse/LIMS-1302)

**Summary**:

UAS often fails to find sessions it has created in the last few minutes, so when loading the second UDC puck from a shipment, Synchweb doesnt assign the session id to the container, even though it knows the session id and so it must exist.

**Changes**:
- Treat a 404 error the same as a success
- Only log errors from UAS, not only successes (UAS.php line 76)
- Fix indentation to use spaces not tabs

**To test**:
- Hard to test without UAS. Currently UAS has a workaround in so it doesn't return a 404, but that will probably be removed soon.
- I suggest (if approved and if UAS take out their workaround) we roll out to staging, then point the beamline to staging temporarily. We can load multiple pucks from one shipment, which should assign the session id, but will log the error from UAS.

### DIFF
--- a/api/src/Page/Proposal.php
+++ b/api/src/Page/Proposal.php
@@ -1198,8 +1198,8 @@ class Proposal extends Page
             $uas = new UAS($auto_user, $auto_pass);
             $code = $uas->update_session($uasSessionId, $data);
 
-            if ($code == 200) {
-                // Update ISPyB records
+            if ($code == 200 || $code == 404) {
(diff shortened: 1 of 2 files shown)


NKatti2011:
Ready to Release (but should test on staging first)

PR #778, opened 2024-05-21
LIMS-1334: Allow selection of plate samples already in a group

ndg63276:
**JIRA ticket**: [LIMS-1334](https://jira.diamond.ac.uk/browse/LIMS-1334)

**Summary**:

Allow a plate sample that is already in a group to be selected for adding to a new group

**Changes**:
- Maintain a list of manually selected samples, pass that down to the plate view
- Use this list to decide whether a click is a select or deselect
- Use uniqBy instead of concatenating a difference list to return valid samples (this list includes those already in groups)

**To test**:
- Open proposal nt37104, go to Sample Groups, then Create Sample Group
- Select a shipment and a plate (with no coloured samples)
- Select a few samples and create a group with a name, check the view on the Sample Groups page that this has worked correctly
- Go back to Create Sample Group, use the same shipment and plate, some samples should now be blue to show they are already in a group
- Check that clicking on blue samples turns them green, and clicking again turns them blue again
- Select some previously blue samples and some previously grey samples and create a group.
- Use the row select, column select, and select all buttons to create more groups. Each should toggle green and blue or grey.
- Check each group has been created with the correct samples.
-

### DIFF
--- a/client/src/js/app/store/modules/store.menus.js
+++ b/client/src/js/app/store/modules/store.menus.js
@@ -7,11 +7,13 @@ import TomoMenu from 'modules/types/tomo/menu.js'
 import XpdfMenu from 'modules/types/xpdf/menu.js'
 import GenProcMenu from 'modules/types/genproc/menu.js'
 import ConexsMenu from 'modules/types/conexs/menu.js'
+import EmMenu from 'modules/types/em/menu.js'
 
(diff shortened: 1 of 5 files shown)


NKatti2011:
Looks good to me

gfrn (approved):

PR #779, opened 2024-05-29
LIMS-1331: Send all dispatch requests unless international via DHL for BI/MX

JPHall-DLS:
**JIRA ticket**: [LIMS-1331: Send all dispatch requests to Goods Handling unless international via DHL for BI/MX](https://jira.diamond.ac.uk/browse/LIMS-1331)

**Summary**:
Revise dewar dispatch logic to send all dispatch requests to Goods Handling unless international via DHL for BI or MX as these require special processing.

**Changes**:
- Query modified to return current proposal code associated with dewar.
- Conditions added to check courier name and proposal code.

**To test**:
In development environment, configure `dispatch_email` and `dispatch_email_intl` options to use different email addresses, the first representing Goods Handling, the latter recipient being responsible for special international shipments. Create new shipment, mark shipment as sent to Diamond, and then submit dispatch requests specifying different courier names and destination countries.

For example, if the following are all true:
1. The courier is DHL
2. The destination is international (i.e. not the `facility_country`)
3. The proposal code is BI or MX (or CM for internal use)

Then a dispatch request message should be sent to `dispatch_email_intl` i.e. recipient responsible for special international shipments.

In all other cases, a dispatch request message should be sent to `dispatch_email` i.e. Goods Handling.


api/src/Page/Shipment.php
```diff
@@ -1122,7 +1122,14 @@ function _dispatch_dewar()
             $data['LCEMAIL'] = '';
         $email->data = $data;
 
-        if ($country != $facility_country && !is_null($dispatch_email_intl)) {
+        if (
+            $country != $facility_country &&
+            array_key_exists('DELIVERYAGENT_AGENTNAME', $data) &&
+            $data['DELIVERYAGENT_AGENTNAME'] === 'DHL' &&
```
ndg63276, line 1128: Does this need to be converted to upper case before comparison?

ndg63276 (approved):

PR #780, opened 2024-06-03
LIMS-1342: Show regular create AWB form if shipping redirect not enabled

ndg63276:
**JIRA ticket**: [LIMS-1342](https://jira.diamond.ac.uk/browse/LIMS-1342)

**Summary**:

[ULIMS-50](https://jira.diamond.ac.uk/browse/ULIMS-50) introduced changes to the create AWB form. These changes should have no effect if 
`$use_shipping_service_redirect_incoming_shipments = False;`
in config.php.

**Changes**:
- Add a condition on `shipping_service_app_url_incoming`, which depends on `$use_shipping_service_redirect_incoming_shipments`

**To test**:
- Set `$use_shipping_service_redirect_incoming_shipments = false;`, then
    - create a shipment from the UK, click Create DHL Air Waybill, check a full form appears, with Get Quote at the bottom
    - click Use Facility Account, check the form stays, but Get Quote changes to Create Air Waybill
    - create a shipment from Aruba, click Create DHL Air Waybill, check a full form appears, with Get Quote at the bottom
    - check the Use Facility Account button is hidden
 - Set `$use_shipping_service_redirect_incoming_shipments = true;` and `$shipping_service_app_url = "something.diamond.ac.uk";`, then
    - create a shipment from the UK, click Create DHL Air Waybill, check a full form appears, and also a dropdown with options for UK or other
    - click Use Facility Account, check the form disappears, leaving only a Proceed button
    - create a shipment from Aruba, click Create DHL Air Waybill, check a full form appears, with Get Quote at the bottom
    - check the Use Facility Account button is hidden

### DIFF
--- a/api/src/Page/Shipment.php
+++ b/api/src/Page/Shipment.php
@@ -1020,10 +1020,11 @@ class Shipment extends Page
         }
 
         $server_port = ($_SERVER['SERVER_PORT']==='443') ? '' : ":{$_SERVER['SERVER_PORT']}";
+        $protocol = isset($_SERVER["HTTPS"]) ? 'https' : 'http';
         $shipment_request_info = array(
             "proposal" => $proposal,
             "external_id" => $external_id,
-            "origin_url" => "https://{$_SERVER['SERVER_NAME']}{$server_port}/shipments/sid/{$shipping_id}",
+            "origin_url" => "{$protocol}://{$_SERVER['SERVER_NAME']}{$server_port}/shipments/sid/{$shipping_id}",
             "packages" => $packages
         );
         $response = $this->shipping_service->create_shipment_request($shipment_request_info);
--- a/api/src/Page/Shipment.php
+++ b/api/src/Page/Shipment.php
@@ -2969,7 +2969,7 @@ class Shipment extends Page
                 // TODO: Use null access operator when we upgrade to PHP 8
                 $error_response = $error_json->content->detail ?? $e->getMessage();
                 $error_status = $error_json->status ? $error_json->status : 400; // Status can be 0
-                $this->_error("Shipping service error: $error_response", $error_status);
+                $this->_error("Shipping service error: " . json_encode($error_response), $error_status);
             }
         }
 
--- a/api/src/Page/Shipment.php
+++ b/api/src/Page/Shipment.php
@@ -1218,7 +1218,7 @@ class Shipment extends Page
                         $error_json = json_decode($e->getMessage());
                         $error_response = $error_json->content->detail ?? $e->getMessage();
                         $error_status = $error_json->status ? $error_json->status : 400; // Status can be 0
-                        $this->_error("Shipping service error: $error_response", $error_status);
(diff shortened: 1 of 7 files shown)


client/src/js/modules/shipment/views/createawb.js
```diff
@@ -138,35 +139,60 @@ define(['backbone',
...
+                        this.ui.submit.text("Proceed")
+                    } else {
+                        this.ui.facc.show()
+                        this.ui.createAWBForm.show()
+                        this.setPickupDetailsRequired(true)
+                        this.setEmailRequired(true)
+                    }
                 } else {
-                    this.ui.facc.show()
+                    this.ui.facc.hide()
```
MattPrit, line 170: In order for shipments that are paid for by the facility, but aren't suitable for the redirect, I think we need something like this. 
```suggestion
                    if (
                        app.options.get('facility_courier_countries').indexOf(this.lc.get('COUNTRY')) > -1 ||
                        app.options.get('facility_courier_countries_nde').indexOf(this.lc.get('COUNTRY')) > -1
                    ) {
                        this.ui.facc.show()
                    } else {
                        this.ui.facc.hide()
                    }
```

MattPrit: That could simplify things. We could then have logic like
```
if (
    app.options.get('facility_courier_countries').indexOf(shipmentCountry) > -1
    && app.options.get("shipping_service_app_url_incoming")
) {
    // Show new form
} else {
    // Show old form
}
```
Where `shipmentCountry` is chosen from a full dropdown.

ndg63276: I couldn't simplify it that much, as the "Use facility account" button visibility depends on a few things. But slightly tidier now.

api/src/Page/Shipment.php
```diff
@@ -1217,7 +1218,7 @@ function _dispatch_dewar()
                         $error_json = json_decode($e->getMessage());
                         $error_response = $error_json->content->detail ?? $e->getMessage();
                         $error_status = $error_json->status ? $error_json->status : 400; // Status can be 0
-                        $this->_error("Shipping service error: $error_response", $error_status);
+                        $this->_error("Shipping service error: " . json_encode($error_response), $error_status);
```
MattPrit, line 1221: Is this change required? If it exists, `$error_json->content->detail` should be a string, if not `$error_reponse` will be whatever string is returned by `$e->getMessage()`.

ndg63276: This is what I get back from `$error_json->content->detail` if I run it from a local machine and forget to put .diamond.ac.uk:
```
[{"loc":["body","origin_url"],"msg":"URL host invalid, top level domain required","type":"value_error.url.host"}]
```

MattPrit: Ah, fair enough. I forgot that validation errors didn't follow the same convention.

MattPrit (approved):
Seems to work as expected. LGTM

PR #784, opened 2024-06-11
LIMS-1170: Dont use thumbnails for crystal snapshots

ndg63276:
Superseding https://github.com/DiamondLightSource/SynchWeb/pull/703

**JIRA ticket:** [LIMS-1170](https://jira.diamond.ac.uk/browse/LIMS-1170)

**Summary:**

Use full-sized crystal snapshot image files, not thumbnails.

Note that I have only touched the crystal snapshot image handling, not the diffraction preview image handling, which also uses a thumbnail image, which is generated upstream by a different mechanism to the snapshot image thumbnails.

**Changes:**

Remove ternary operation to switch between reading the full-sized image and the thumbnail. Always use the full-sized image.
I have removed the option to use the /f/1 URL parameter as it is no longer needed, and thus removed it every time it is used.

**To test:**

Find a data collection that has multiple image snapshots (each DC can have up to four).

- Open all views that display a data collection pane. Check that the snapshot image is displayed correctly.
- Click on the snapshot image. Check that you can still roll through the full selection of available snapshot images.
- Repeat the above for a robot load pane.
- Repeat the above for a sample action (eg anneal, wash) - see mx34438-111 for examples
- Repeat the above for grid scans, click the arrows icon to view full size and scroll through all images

@benjaminhwilliams, can you check you are happy with this?


benjaminhwilliams (approved):
Thanks @ndg63276, I'd completely forgotten about my change set on an unmerged branch.  Thanks for picking up where I left off.  LGTM.

PR #792, opened 2024-06-26
LIMS-1355: Make OAuth2 support more generic

gfrn:
**JIRA ticket**: [LIMS-1355](https://jira.diamond.ac.uk/browse/LIMS-1355)

**Summary**:

Currently, support for OAuth2 code flow in the OIDC authentication type is tailored to CAS. In order to enable Keycloak compatibility, it must be configurable, down to the claim used for identifying users, as well as ensuring the whole process is not specific to one or other CAS quirk.

*NOTE:* This requires configuration changes to SynchWeb when being deployed. Do not merge this in before altering the configuration to provide a full path (including protocol) for `$sso_url` and adding `$sso_user_key` to the configuration file with the value `id`.

**Changes**:
- Added new configuration key, `$sso_user_key`, allowing deployer to set which one of the claims returned from the upstream auth service will be used for identification (example: `fedid`)
- `$sso_url` now takes the full absolute path to the endpoint descriptor endpoint in the upstream authentication service
- Internal changes to code flow to support Keycloak and other SSO providers

**To test**:
- Add valid client ID/secret to config, set `$cas_sso` to true, set the authentication type to OIDC, and try to log in with either CAS or Keycloak
- Go to a protected page (e.g.: `proposals`) and try to log in, verify that the redirect occurs

### DIFF
--- a/api/config_sample.php
+++ b/api/config_sample.php
@@ -32,7 +32,9 @@
 
     # Follow CAS SSO
     $cas_sso = true;
-    $sso_url = "sso.server.ac.uk";
+    $sso_url = "https://sso.server.ac.uk";
+    # Profile field to use to identify user
+    $sso_user_key = "fedid";
 
     # OIDC (or OAuth2) client ID and secret. Only useful if authentication_type is set to OIDC
     $oidc_client_id = "oidcClientId";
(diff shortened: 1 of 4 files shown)


ndg63276 (approved):
Works for me

PR #793, opened 2024-06-27
LIMS-1341: Allow entry of SMILES codes

ndg63276:
**JIRA ticket**: [LIMS-1341](https://jira.diamond.ac.uk/browse/LIMS-1341)

**Summary**:

Add an editable box for SMILES code. Regex is fun as lots of characters are allowed, see https://en.wikipedia.org/wiki/Simplified_molecular-input_line-entry_system

**Changes**:
- Add entry box to plate sample page
- Add code to allow editing in the database, and get current value

**To test**:
- Go to proposal nt37104, go to Containers, click on a plate
- Check the SMILES code box appears under sample
- Edit the SMILES code, using weird symbols eg `.:%=#$@+-[]()/\`, check it remains after a refresh
- Change sample, check the value changes as expected.


gfrn:
All tests pass! However, there is no indication to the user that a name is invalid, that is, the `PATCH` request goes through fine, but the `SMILES` field gets ignored in the backend. Would it be useful to add a frontend check as well, or return a 400/422 from the backend if the SMILES code is bad?

gfrn (approved):

PR #804, opened 2024-07-15
LIMS-1370: Handle errors from the shipping service

ndg63276:
**JIRA ticket**: [LIMS-1370](https://jira.diamond.ac.uk/browse/LIMS-1370)

**Summary**:

With `$use_shipping_service = True;` but `$use_shipping_service_redirect = False;`, if the shipping service returns an error, we should display the message detail in Synchweb.

**Changes**:
- Dont try and get the AWB number if there was an error from the shipping service
- Pass the error message back to the GUI as a JSON encoded string, so it appears in the GUI

**To test**:
- NB this is a user dewar so only test on backup db, without sending emails
- Set `$use_shipping_service = True;` but `$use_shipping_service_redirect = False;`
- Go to a shipment that has already been dispatched by the shipping service, eg /shipments/sid/60942
- Click on the little house icon to request the dewar be returned, click "Request Dewar Dispatch"
- Check if a "friendly" message appears at the top of the page, saying "Cannot update a shipment that has already been dispatched"
- Go to a shipment that has not been dispatched eg /shipments/sid/63609, repeat the test, check it all works this time and appears on sample-shipping-staging.

### DIFF
--- a/api/src/Page/Shipment.php
+++ b/api/src/Page/Shipment.php
@@ -1113,7 +1113,7 @@ class Shipment extends Page
             $shipment_id = $response['shipmentId'];
             $this->shipping_service->dispatch_shipment($shipment_id, false);
         } catch (Exception $e) {
-            throw new Exception("Error returned from shipping service: " . $e . "\nShipment data: " . json_encode($shipment_data));
+            throw new Exception($e->getMessage());
         }
 
         return $shipment_id;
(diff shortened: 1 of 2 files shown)


api/src/Page/Shipment.php
```diff
@@ -1226,12 +1226,14 @@ function _dispatch_dewar()
...
+                            $data['AWBURL'] = $this->shipping_service->get_awb_pdf_url($shipment_id);
+                        }
                     } catch (Exception $e) {
-                        error_log($e);
-                        $data['AWBURL'] = "";
-                    }
-                    if (Utils::getValueOrDefault($shipping_service_links_in_emails)) {
-                        $data['AWBURL'] = $this->shipping_service->get_awb_pdf_url($shipment_id);
+                        $error_json = json_decode($e->getMessage());
+                        $error_response = $error_json->content->detail;
```
MattPrit, line 1234:
```suggestion
                        $error_response = $error_json->content->detail ?? $e->getMessage();
```
As at line 1219 (just in case)

api/src/Page/Shipment.php
```diff
@@ -1226,12 +1226,14 @@ function _dispatch_dewar()
                 } else {
                     try {
                         $shipment_id = $this->_dispatch_dewar_in_shipping_service($data, $dew);
+                        if (Utils::getValueOrDefault($shipping_service_links_in_emails)) {
+                            $data['AWBURL'] = $this->shipping_service->get_awb_pdf_url($shipment_id);
+                        }
                     } catch (Exception $e) {
-                        error_log($e);
```
MattPrit, line 1230: Do we still want to log the exception?

MattPrit (approved):

MattPrit (approved):

Issue #552, opened 2023-05-11
Bulk sample update is difficult to use


ejd53:
Chrome, Version 112.0.5615.140 (Official Build) (64-bit)

In the Container view, e.g., https://ispyb.diamond.ac.uk/containers/cid/260277, there is a table of samples within the puck. There is a bulk-populate form above this table. It has some usability issues.

**Possible to update in intermediate state**

There can be a delay of several seconds between selecting a column and having it highlighted and populated with the new values. This is easily enough time to select a column, select a value, and hit Update Samples. It is possible to submit the update form during this delay, but it does not update the underlying records, despite the "successfully updated" message appearing. I assume that it's "updating" with the values that are in the table when the AJAX request is sent, rather than with those selected in the bulk update form.

It's also possible to change and submit the bulk update form after the AJAX update has been sent, but before it completes, further adding to the confusion.

Ideally, the table would be populated with the updated values so quickly as to render it impossible to submit the form before this happens. Failing that, it should not be possible to submit the form until the table updates, and certainly not while the AJAX update is in progress. It'd be nice to have some visual indication, whether a spinner or just disabling the form, to show that UI and AJAX updates are in progress.

**Singular v plural**

The success message refers to "sample", singular, when all samples in the puck were (or should have been) updated.

**Insufficiently clear notification of errors**

"Successfully updated" appears but the fields are not updated, and it's not obvious that there was an error. For example, "Screening Method" can't be set to "better than" without a minimum resolution value having been set:

To reproduce:
Open the Unattended Data Collection tab of the table. Note that "Minimum resolution" column contains no values.
Select "Screening Method" and "Better than" in the bulk update form.
Wait for table fields to be highlighted.
Click "Update samples".

Expected: 
Fields are saved.
Green "successfully updated" bar appears at top of window.
(or)
Fields are not saved
Red error message appears at top of window
Green "successfully updated" does not appear

Actual:
Fields are not saved.
Green "successfully updated" bar appears at top of window.
"Update samples" button gains a thin red border.
Red "minimum resolution must be set" errors appear out of sight below table.
Error box disappears, often before you can scroll down to read it (if you know it's there)

ndg63276:
Oddly, I get different behaviour for point number 3. I don't get the green "successfully updated" bar at the top, and the errors at the bottom don't disappear, even if I wait >30s. Using Chrome (113) and Firefox (102.8).

[...]

PR #808, opened 2024-07-18
LIMS-700: Allow bulk editing of protein

ndg63276:
**JIRA ticket**: [LIMS-700](https://jira.diamond.ac.uk/browse/LIMS-700)
**JIRA ticket**: [LIMS-691](https://jira.diamond.ac.uk/browse/LIMS-691)
**Github issue**: https://github.com/DiamondLightSource/SynchWeb/issues/552

**Summary**:

Improve the behaviour of the bulk editor.

**Changes**:
- Disable the bulk editor dropdown, textbox and submit button until the previous request has finished
- Shorten the debounce time on updating the samples to make the form respond quicker
- Add (green only) protein acronym to the fields that can be bulk edited
- Sort the fields alphabetically by name rather than "field name"
- Correct the message displayed when samples are updated successfully
- Add an error notification if samples are not updated

**To test**:
- Make a new puck with some samples in, save it, and then view it
- Check the bulk editor appears with the dropdown in alphabetical order (ie Barcode before Cell A)
- Select Comment and type a new comment, check it appears promptly next to each sample
- Open the dev tools, click the Update Samples button a few times, check only one PUT request is sent
- Select Protein Acronym, choose a new acronym, click Update Samples, then refresh the page to check the change has persisted
- Check the success message appears and says "Sample(s) have been successfully updated"
- Choose "Screening Method", then select "Better Than", but dont choose a minimum resolution. Click Update Samples, and check a notification appears at the top to say there was an error, and validation messages appear at the bottom.
- Go to proposal mx31800, find a container to bulk edit (or make a new one), choose Protein Acronym, check that Diazocine does not appear in the list as it is a YELLOW sample
- Go to a container that has had data collected on it, check the bulk editor is hidden

### DIFF
--- a/client/src/js/modules/types/mx/samples/sample-table-mixin.js
+++ b/client/src/js/modules/types/mx/samples/sample-table-mixin.js
@@ -52,6 +52,16 @@ export default {
           SAFETYLEVEL: item.SAFETYLEVEL
         }));
     },
+    greenProteinsList() {
+      return this.$proteins()
+        .filter(item =>
+          item.SAFETYLEVEL === "GREEN"
+        )
+        .map(item => ({
+          value: item.PROTEINID,
+          text: item.ACRONYM,
+        }));
+    },
     experimentKindList() {
       return this.$experimentKindList();
     },
--- a/client/src/js/modules/types/mx/samples/mx-puck-samples-table.vue
+++ b/client/src/js/modules/types/mx/samples/mx-puck-samples-table.vue
@@ -30,6 +30,7 @@
           option-value-key="key"
           :options="mappedEditableColumns"
           :value="selectedEditableColumn.key"
+          :isDisabled="disableUpdateSamples"
           outer-class="tw-mr-2"
           input-class="tw-h-8"
           @input="resetSelectedEditableColumn"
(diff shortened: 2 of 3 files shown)


client/src/js/modules/types/mx/samples/sample-table-mixin.js
```diff
@@ -52,6 +52,16 @@ export default {
           SAFETYLEVEL: item.SAFETYLEVEL
         }));
     },
+    greenProteinsList() {
+      return this.$proteins()
+        .filter(item =>
```
gfrn, line 57: Is it worth it performing this check in the backend too? Just in case someone tries to "push through anyway"

(On a slightly less useful note, I think we can use `reduce` to perform this in O(N) (N operations, N being the array length), but that's just being pedantic and I don't think we gain much performance anyway)

ndg63276: I wasn't sure if any other function was using the backend function, so didnt want to change it to stop Yellow/Red proteins. It also does each sample in turn on the backend, so you get a situation where some might succeed and then others fail.

gfrn: That makes sense! And it should be similar to changing it manually, so it's not a major risk to my eyes

gfrn (approved):

PR #817, opened 2024-08-20
LIMS-656: Show auto subsamples if they have data

ndg63276:
**JIRA ticket**: [LIMS-656](https://jira.diamond.ac.uk/browse/LIMS-656)

**Summary**:

If a subsample on a plate has an associated data collection, then it should always be shown, even if it is from an old inspection.

**Changes**:
- Add an optional extra UNION to the subsample query showing "auto" samples with data collections

**To test**:
- Go to /containers/cid/249257 and click on the bottom left most well (well 169). There should be (at time of writing) 4 extra subsamples in the list under "Marked Subsamples", the first has X=2286, Y=716 and corresponds to /dc/visit/nt30330-87/id/10216910
image

- Check well 1 on the same plate, there should be no difference in the number of subsamples shown as none have any data


gfrn:
Well 1 seems to have one extra subsample (10 items are shown in the list), is that because new data was inserted since this ticket was created?

As an aside, clicking the "view collected data" button on a subsample, then clicking the permalink button, leads to an oddly placed modal being present, although this has no effect on the changes presented in this ticket.

ndg63276:
> Well 1 seems to have one extra subsample (10 items are shown in the list), is that because new data was inserted since this ticket was created?

Sorry, I didn't word it very well. Well 1 should have 10 items before and after this change. Well 169 should have 5 items before, and 9 items after this change.

gfrn (approved):

PR #826, opened 2024-09-05
LIMS-1354: Migrate from ActiveMQ to RabbitMQ

JPHall-DLS:
**JIRA ticket**: [LIMS-1354](https://jira.diamond.ac.uk/browse/LIMS-1354)

**Summary**:
SynchWeb sends messages to ActiveMQ to trigger MX reprocessing and AlphaFold model prediction jobs. A single ActiveMQ server proved unreliable so a RabbitMQ cluster was established in early 2021. An ActiveMQ to RabbitMQ bridge service was implemented as a temporary solution while applications migrated to RabbitMQ. The failure of the bridge service in mid-2024 highlighted the need to update SynchWeb to send messages direct to RabbitMQ.
SynchWeb uses the Stomp (Simple Text Oriented Message Protocol) PHP client to send messages to ActiveMQ. This ticket updates SynchWeb to use an AMQP (Advanced Message Queuing Protocol) client to send messages to RabbitMQ. The php-amqplib client is the de facto standard and detailed in the RabbitMQ PHP tutorial.

**Changes**:
- Update composer.json to replace stomp-php dependency with php-amqplib.
- Add global variables for RabbitMQ configuration that reflect RabbitMQ nomenclature.
- Update Queue class and Queue::send method to use AMQPStreamConnection and AMQPMessage. Remove redundant ActiveMQ header information e.g. synchweb.host and synchweb.user.
- Update Process class to use new global variable names.
- Update Page class to use new global variable names, Queue class signature, and Queue::send method signature.

**To test**:
- Reprocess MX data via the gearwheel icon on the Data Collection page. In the database, SynchWeb will create a new record in the ProcessingJob table. SynchWeb will send a message to RabbitMQ with this ProcessingJobId. Zocalo will read the message from RabbitMQ and create a corresponding record in the AutoProcProgram table. A new processing job will soon be shown under Auto Processing on the Data Collection page.
- AlphaFold model prediction jobs are triggered when proteins are added or updated with a sequence. Model prediction does not currently work as AlphaFold itself was compressed in the great GPFS02 purge of 2024.

### DIFF
--- a/api/src/Queue.php
+++ b/api/src/Queue.php
@@ -19,22 +19,16 @@ class Queue
 
     function send($vhost, array $message)
     {
-        try {
-            $connection = new AMQPStreamConnection($this->host, $this->port, $this->username, $this->password);
-            $channel = $connection->channel();
+        $connection = new AMQPStreamConnection($this->host, $this->port, $this->username, $this->password);
+        $channel = $connection->channel();
 
-            $msg = new AMQPMessage(
-                json_encode($message, JSON_UNESCAPED_SLASHES)
-            );
+        $msg = new AMQPMessage(
+            json_encode($message, JSON_UNESCAPED_SLASHES)
+        );
 
-            $channel->basic_publish($msg, '', $vhost);
+        $channel->basic_publish($msg, '', $vhost);
 
-            $channel->close();
-            $connection->close();
-        } catch (AMQPException $e) {
-            /** @noinspection PhpUnhandledExceptionInspection */
-
-            throw $e;
-        }
+        $channel->close();
+        $connection->close();
     }
 }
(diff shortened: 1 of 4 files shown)


api/src/Queue.php
```diff
@@ -2,39 +2,36 @@
...
+                json_encode($message, JSON_UNESCAPED_SLASHES)
             );
 
-            $connection->disconnect();
-        } catch (StompException $e) {
+            $channel->basic_publish($msg, '', $vhost);
+
+            $channel->close();
+            $connection->close();
+        } catch (AMQPException $e) {
```
ndg63276, line 34: AMQPException is not imported. But if we are just going to throw the exception, we could probably just get rid of the try block?

ndg63276 (approved):

PR #827, opened 2024-09-10
LIMS-1099: Limit what can be done with red proteins

ndg63276:
**JIRA ticket**: [LIMS-1099](https://jira.diamond.ac.uk/browse/LIMS-1099)

**Summary**:

We should not allow red/yellow proteins into green shipments, nor allow containers with them to be queued for UDC.

**Changes**:
- Shipments must either be associated with a visit, or tick the UDC or Responsive box (to allow "-")
- If a shipment is created as Red or Yellow:
  - hide the UDC and Responsive boxes, so a visit must be chosen
  - hide the "Queue for UDC" button
- Only allow proteins in a suitable shipment, eg no red proteins in a green shipment
- Don't allow changing shipment safety level to lower than that of the proteins
- Don't allow changing shipment safety level to yellow/red if any of the containers are queued for UDC

**To test**:
- Go to a proposal with future visits eg cm37235
- Go to the Add Shipment page. Check you cannot make a shipment without selecting a First Experiment.
- Change the Safety Level to yellow or red, check the UDC / Imager and Responsive / Mail in checkboxes are hidden.
- Choose a First Experiment and check you can make a shipment
- Add a container, check the Queue For UDC checkbox (while adding) or button (while viewing after) is hidden as the shipment is yellow or red
- Edit the shipment so it is green, add another container, check the Queue For UDC checkbox is shown, and you can queue the first container as well now
- Try to edit the shipment so that it is yellow again, check an error message appears saying you cannot as containers are queued
- Go to a proposal with yellow proteins eg mx37045
- Go to the Add Shipment page, make a green shipment, ticking the UDC checkbox
- Add a container, check the yellow protein NS_PGO is not listed.
- Go back to the shipment, change it to a yellow shipment. Edit the container, check that NS_PGO is now listed, and select that for at least one sample.
- Go back to the shipment, try to change the shipment back to green, check an error message appears saying you cannot as some proteins are yellow. Check you can still change the shipment to red.

### DIFF
--- a/api/src/Page/Sample.php
+++ b/api/src/Page/Sample.php
@@ -1805,12 +1805,17 @@ class Sample extends Page
             $where .= ' AND pr.externalid IS NOT NULL';
         }
 
-        $has_safety_level = $this->has_arg('SAFETYLEVEL');
-        if ($has_safety_level && $this->arg('SAFETYLEVEL') === 'ALL') {
-            $where .= ' AND pr.safetyLevel IS NOT NULL';
-        } else if ($has_safety_level && $this->arg('SAFETYLEVEL') !== 'ALL') {
-            $where .= ' AND pr.safetyLevel=:' . (sizeof($args) + 1);
-            array_push($args, $this->arg('SAFETYLEVEL'));
+        if ($this->has_arg('SAFETYLEVEL')) {
+            if ($this->arg('SAFETYLEVEL') === 'ALL') {
+                $where .= ' AND pr.safetyLevel IS NOT NULL';
+            } else if (strtolower($this->arg('SAFETYLEVEL')) === 'yellow') {
+                $where .= " AND pr.safetyLevel in ('green','yellow')";
+            } else if (strtolower($this->arg('SAFETYLEVEL')) === 'red') {
+                $where .= " AND pr.safetyLevel in ('green','yellow','red')";
+            } else {
+                $where .= ' AND pr.safetyLevel=:' . (sizeof($args) + 1);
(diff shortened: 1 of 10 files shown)


gfrn:
It would be nice if when you attempt to change a shipment safety level to "yellow" with queued containers, it reverted to "green" after clicking OK. This is in complement to the warning message, since users might not see it as it fades out.

gfrn:
When creating a sample from a yellow protein in a container in a yellow shipment (such as NS_PGO in mx37045), the POST request for sample creation goes through, return 200, but doesn't seem to create any samples. In fact, an empty array is returned as the response body

Is this intended?

ndg63276:
> It would be nice if when you attempt to change a shipment safety level to "yellow" with queued containers, it reverted to "green" after clicking OK. This is in complement to the warning message, since users might not see it as it fades out.

Yep, good point. Have added a new option to the Editable class to revert the value displayed if the server gives an error. So now you get the warning at the top, and the value reverts as well.

gfrn:
> > When creating a sample from a yellow protein in a container in a yellow shipment (such as NS_PGO in mx37045), the POST request for sample creation goes through, return 200, but doesn't seem to create any samples. In fact, an empty array is returned as the response body
> > Is this intended?
> 
> That's weird, it creates samples for me, and the response body is an array of the created samples.

Seems like this is part of a recurring issue I mentioned in other PR, it started working this morning, even though I cleared data yesterday a couple of times. That's definitely odd, and possibly a bit of undefined behaviour

gfrn (approved):

PR #834, opened 2024-09-18
LIMS-1286: Show ERA status on calendar and visits list

ndg63276:
**JIRA ticket**: [LIMS-1286](https://jira.diamond.ac.uk/browse/LIMS-1286)

**Summary**:

Users (and staff) often don't realise their ERA has not been validated as the visit approaches, so we should display it more prominently.

**Changes**:
- Get risk rating for each session from ISPyB (it should be synced to UAS)
- Display a green/yellow/red circle for low/medium/high risk sessions, and a black circle if the ERA has not been approved yet, on both the calendar page and also the visits list page
- Display a tooltip to explain the circle if a user hovers over it

**To test**:
- Go to an MX proposal (eg /dc/visit/mx37045-1) and then click the calendar link at the top left. As staff you should see lots of visits for the mx village, check a coloured circle appears next to each, and a tooltip appears if you hover over it. Check there are some green, yellow and black.
- Go to the /visits page, check there is an ERA column with the same circles and tooltip.
- Go to proposal mx18802 and check the visits list, this proposal has High risk sessions from May 2018, so should have red circles.

### DIFF
--- a/api/src/Page/Proposal.php
+++ b/api/src/Page/Proposal.php
@@ -456,6 +456,7 @@ class Proposal extends Page
                     s.beamlineoperator                                            AS lc,
                     s.comments,
                     s.scheduled,
+                    s.riskrating,
                     st.typename                                                   AS sessiontype,
                     DATE_FORMAT(s.startdate, '%d-%m-%Y %H:%i')                    AS startdate,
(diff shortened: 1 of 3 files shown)


NKatti2011 (approved):
LGTM

PR #841, opened 2024-09-30
Improvement/lims 1458/fix GitHub warnings

ndg63276:
**JIRA ticket**: [LIMS-1458](https://jira.diamond.ac.uk/browse/LIMS-1458)

**Summary**:

We should try to decrease the number of JavaScript errors in the "JavaScript build, test and lint" job that runs on creation of a PR.

**Changes**:
- Change `this.attributes.FOO` to `this.get('FOO')`
- Change `this.attributes.FOO = 'BAR'` to `this.set('FOO', 'BAR')`
- Move declaration of or rename js variables so that they are not redefined
- Remove unnecessary escape characters eg `\:`
- Add a log message to avoid having an empty catch statement

**Not done**
- `Use this.$ instead of $ in views` - changing seems to break functionality
- `Do not silence events` - not sure what this does
- `Do not access Object.prototype method 'hasOwnProperty' from target object` - not sure how to fix this

**To test**:
- Manually check the code :(
- Check basic Synchweb functions as expected
- Click "Show all checks" in the "All checks have passed" message below, then click Details on the JavaScript job
- Click on the "JavaScript build, test and lint" task, scroll down and see if there are less than about 50 errors.

### DIFF
--- a/client/src/js/modules/cell/views/beamlines.js
+++ b/client/src/js/modules/cell/views/beamlines.js
@@ -9,6 +9,11 @@ define(['marionette', 'utils',
         template: false,
         modelEvents: { 'change': 'render' },
 
+        ascsort: function(a,b) {
+            if (a[0] == b[0]) return 0
+            else return (a[0] < b[0]) ? -1 : 1
+        },
+
         onRender: function() {
               if (this.model.get('data')) {
                   var did = this.getOption('second') ? 1 : 0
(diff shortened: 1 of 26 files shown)


client/src/js/modules/cell/views/beamlines.js
```diff
@@ -9,6 +9,11 @@ define(['marionette', 'utils',
         template: false,
         modelEvents: { 'change': 'render' },
 
+        ascsort: function(a,b) {
```
gfrn, line 12: Since this is a static function, and does not depend on the current context, could this be moved to an utils file?

Otherwise, would `.sort().reverse()` work? It is slightly less efficient, however

client/src/js/modules/dc/views/reprocess2.js
```diff
@@ -156,8 +156,8 @@ define(['backbone', 'marionette', 'views/dialog',
             if (e) e.preventDefault()
 
             if (this.aps.length) {
-                var e = this.aps.at(0)
-                var c = e.get('CELL')
+                var a = this.aps.at(0)
+                var c = a.get('CELL')
```
gfrn, line 160: Might be a good opportunity to switch these out for `let` or `const` instead, which does not allow in-scope redeclaration of variables

```suggestion
                const a = this.aps.at(0)
                const c = a.get('CELL')
```

client/src/js/modules/cell/views/beamlines.js
ndg63276: I decided it was easier to have the database do the sorting. I had some trouble with the plotting library not liking undefined combinations of year and beamline, so added some code to set them all to 0 and then replacing with the real count.

To test:
* Go to /statistics/pdb
* Check the stacked bar charts labelled "Depositions by beamline / year" look ok (up to 2020 when it was last updated)

gfrn (approved):

PR #852, opened 2024-10-29
LIMS-1490: Use callback URL for incoming dewars via shipping service

ndg63276:
**JIRA ticket**: [LIMS-1490](https://jira.diamond.ac.uk/browse/LIMS-1490)

**Summary**:

If we are to fully use the shipping service for incoming dewars, we need to use the callback URL function to allow setting of the dewar status and tracking number, once the shipping service booking is complete.

**Changes**:
- Add a new (non-authenticated) endpoint (`/api/shipment/dewars/confirmpickup/sid/<shippingid>/token/<token>`) for the shipping service to hit when an incoming dewar has been booked
  - Update the shipping status & dewar status to `awb created` or `pickup booked` as necessary
  - Update the dewar tracking number
  - Add lines to the DewarTransportHistory table
- Use an md5 token to authenticate the new endpoint
- Hide irrelevant fields on the Shipment page if the shipping service has been used

**To test**:
- The dev instance needs to be on a *.diamond.ac.uk address, so use a beamline workstation or deploy to ispyb-dev-*
- Set these variables in config.php
```
$use_shipping_service = True;
$use_shipping_service_incoming_shipments = True;
$shipping_service_api_url = "https://sample-shipping-test.diamond.ac.uk/api";
$shipping_service_api_user = "synchweb";
$shipping_service_api_password = <redacted>;
$shipping_service_app_url = "https://sample-shipping-test.diamond.ac.uk";
$use_shipping_service_redirect_incoming_shipments = True;
$use_shipping_service_redirect = True;
```
- Create a shipment with more than one dewar in a UK proposal, eg mx23694
- On the shipment page, click Create DHL Air Waybill
- Tick the boxes for both dewars, click Use Facility Account and accept the terms
- Click Proceed and you should be redirected to the Shipping Service
- Fill in the address and request a pickup for the following day
- Exit back to Synchweb, and:
  - check the Shipment status and both dewars' status is 'pickup booked'
  - check the "Tracking # to" has been filled for both dewars
  - check the history section has both "awb created" and "pickup booked"
  - check lots of the fields on the Shipment page have been hidden, eg Courier, Courier Account No etc
- Rerun the previous test but don't request a pickup in the shipping service (untick the box). The shipping/dewar status should stay at "awb created" and the history should only have "awb created"
- Set `$use_shipping_service_redirect_incoming_shipments = False;`, create and book a shipment without being redirected to the shipping service, check that still works. Check the Courier, Courier Account No etc fields are displayed.

### DIFF
--- a/api/src/Controllers/AuthenticationController.php
+++ b/api/src/Controllers/AuthenticationController.php
@@ -119,8 +119,7 @@ class AuthenticationController
             ($parts[0] == 'shipment' && $parts[1] == 'containers' && $parts[2] == 'history' && in_array($_SERVER["REMOTE_ADDR"], $bcr)) ||
 
                 # Allow shipping service to update dewar status
-            ($parts[0] == 'shipment' && $parts[1] == 'dewars' && $parts[2] == 'confirmdispatch') ||
-            ($parts[0] == 'shipment' && $parts[1] == 'dewars' && $parts[2] == 'confirmpickup')
+            ($parts[0] == 'shipment' && $parts[1] == 'dewars' && ($parts[2] == 'confirmdispatch' || $parts[2] == 'confirmpickup'))
             )
             {
                 $need_auth = false;
--- a/api/src/Controllers/AuthenticationController.php
+++ b/api/src/Controllers/AuthenticationController.php
@@ -119,7 +119,8 @@ class AuthenticationController
             ($parts[0] == 'shipment' && $parts[1] == 'containers' && $parts[2] == 'history' && in_array($_SERVER["REMOTE_ADDR"], $bcr)) ||
 
                 # Allow shipping service to update dewar status
-            ($parts[0] == 'shipment' && $parts[1] == 'dewars' && $parts[2] == 'confirmdispatch')
+            ($parts[0] == 'shipment' && $parts[1] == 'dewars' && $parts[2] == 'confirmdispatch') ||
+            ($parts[0] == 'shipment' && $parts[1] == 'dewars' && $parts[2] == 'confirmpickup')
             )
             {
                 $need_auth = false;
--- a/api/src/Page/Shipment.php
+++ b/api/src/Page/Shipment.php
@@ -146,6 +146,7 @@ class Shipment extends Page
         'TOKEN' => '\w+',
         'tracking_number' => '\w+',
         'AWBURL' => '[\w\:\/\.\-]+',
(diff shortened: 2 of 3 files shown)


api/src/Controllers/AuthenticationController.php
```diff
@@ -119,7 +119,8 @@ private function checkAuthRequiredForSpecificSituations($parts): bool
             ($parts[0] == 'shipment' && $parts[1] == 'containers' && $parts[2] == 'history' && in_array($_SERVER["REMOTE_ADDR"], $bcr)) ||
 
                 # Allow shipping service to update dewar status
-            ($parts[0] == 'shipment' && $parts[1] == 'dewars' && $parts[2] == 'confirmdispatch')
+            ($parts[0] == 'shipment' && $parts[1] == 'dewars' && $parts[2] == 'confirmdispatch') ||
+            ($parts[0] == 'shipment' && $parts[1] == 'dewars' && $parts[2] == 'confirmpickup')
```
gfrn, line 123: This might be slightly cleaner

```suggestion
            ($parts[0] == 'shipment' && $parts[1] == 'dewars' && ($parts[2] == 'confirmdispatch' || $parts[2] == 'confirmpickup'))
```

api/src/Page/Shipment.php
```diff
@@ -3058,20 +3123,27 @@ function($package, $index) {return array("piecenumber" => $index+1, "licenseplat
     function _create_shipment_shipment_request($shipment, array $dewars): int
     {
 
-        // if (!is_null($shipment['EXTERNALSHIPPINGIDTOSYNCHROTRON'])) {
-        //     return $shipment['EXTERNALSHIPPINGIDTOSYNCHROTRON'];
-        // }
+        $shipping_id = (int) $shipment['SHIPPINGID'];
+
+        $token = md5(uniqid());
```
gfrn, line 3128: Bit of a nitpick, but `uniqid` could be "guessed", since we're MD5 hashing it rather than encoding it with an asymmetric cryptographic key pair

In this case, I think `openssl_random_pseudo_bytes` might work better:

```suggestion
        $token = md5(openssl_random_pseudo_bytes(7));
```

(This still isn't the safest, since MD5 hashes have been know to be subject to collisions for a while now, but I guess it's in tune with PHP's docs!)

gfrn (approved):

PR #853, opened 2024-11-01
LIMS-1515: Add energy value to data collections

ndg63276:
**JIRA ticket**: [LIMS-1515](https://jira.diamond.ac.uk/browse/LIMS-1515)

**Summary**:

Currently only wavelength is displayed in Synchweb for most data collection types, we should show energy in eV as well.

**Changes**:
- Define the eV/A conversion factor (value from https://en.wikipedia.org/wiki/Electronvolt#Wavelength)
- Add return values to the `_data_collections` function for the `energy` in eV, and also the `wavelength` in A for MCA spectra and peak (`wpk`) and inflection (`win`) wavelengths for edge scans
- Add labels to a few more columns, these are redundant as the union will use the labels from the first select statement, but help to identify where values are coming from
- Return energies to 0 decimal places and wavelengths to 4 decimal places, except for edge scan results which have 1 and 5 respectively (these are results so more precision is needed)
- Display these values, taking conversion factor out of the JS
- Make the edge scan filename a double width column, so that E(peak) is displayed next to the peak f''/f' results, and the same for E(inflection). Cut all f''/f' values to 2 decimal places.

**To test**:
- View a standard data collection (eg /dc/visit/cm37235-4/id/15542386), check the wavelength and energy are displayed as `Wavelength: 0.9537Å (13,000eV)`
- View a grid scan (eg /dc/visit/cm37235-4/id/15542377), check the wavelength and energy are displayed as `Wavelength: 0.9537Å (13,000eV)`
- View an MCA spectrum (eg /dc/visit/cm37235-4/ty/mca/id/15813), check the energy is displayed as `Energy: 18,000eV (0.6888Å)`
- View an edge scan (eg /dc/visit/cm37235-4/ty/edge/id/61197), check:
  - energies have 1 decimal place, wavelengths have 5 decimal places
  - f''/f' values have 2 decimal places, and e- as units
  - peak f''/f' is displayed next to peak energy, inflection f''/f' is displayed next to inflection energy, ie
image

### DIFF
--- a/api/src/Page/DC.php
+++ b/api/src/Page/DC.php
@@ -62,7 +62,7 @@ class DC extends Page
         array('/dat/:id', 'get', '_plot'),
     );
 
-    public $eVToA = 12398.4198;
+    const EVTOA = 12398.4198;
 
     # ------------------------------------------------------------------------
     # Data Collection AJAX Requests
@@ -424,7 +424,7 @@ class DC extends Page
                     dc.transmission,
                     dc.axisrange,
                     dc.wavelength,
-                    ".$this->eVToA."/dc.wavelength as energy,
+                    ".self::EVTOA."/dc.wavelength as energy,
                     dc.comments,
                     1 as epk,
                     1 as ein,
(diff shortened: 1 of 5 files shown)


api/src/Page/DC.php
```diff
@@ -62,6 +62,8 @@ class DC extends Page
         array('/dat/:id', 'get', '_plot'),
     );
 
+    public $eVToA = 12398.4198;
```
gfrn, line 65:
```suggestion
    public const $eVToA = 12398.4198;
```

Since we're using it in a SQL query in a position that is rather sensitive to injections, it's better to explicitly mark it as a constant

ndg63276: That gave an error, but I've done it as a class constant in the commit below.

gfrn (approved):

PR #854, opened 2024-11-01
LIMS-1514: Allow sorting of visit list

ndg63276:
**JIRA ticket**: [LIMS-1514](https://jira.diamond.ac.uk/browse/LIMS-1514)

**Summary**:

The list of visits for a proposal is not sortable.

**Changes**:
- Combine the data collection count into the main db query so that it can be sorted on
- Remove the need for an enter press on searching for a visit number, but add a debounce
- Make column headers actually be buttons so that the cursor becomes a pointer
- Re-fetch the data with a sort parameter if a header is clicked (apart from "Dewars")
- Re-fetch the data in the opposite direction if the same header is clicked again
- Don't add a search parameter if the search box is empty
- Allow searching by beamline name

**To test**:
- Open a proposal with a bunch of visits eg mx34263, and go to the /visits page
- Check you can sort by each header, in both directions, apart from "Dewars"
- Check the number of Data Collections is the same as on prod
- Search for a visit number, check only that visit number is displayed, check the debounce is working by looking at the requests sent
- Search for a beamline (eg i03), check only visits for that beamline are displayed

### DIFF
--- a/api/src/Page/Proposal.php
+++ b/api/src/Page/Proposal.php
@@ -393,7 +393,8 @@ class Proposal extends Page
         }
 
         if ($this->has_arg('s')) {
-            $where .= " AND s.visit_number LIKE :" . (sizeof($args) + 1);
+            $where .= " AND (s.visit_number LIKE :" . (sizeof($args) + 1) . " OR s.beamlinename LIKE :" . (sizeof($args) + 2) . ")";
+            array_push($args, $this->arg('s'));
             array_push($args, $this->arg('s'));
         }
 
--- a/api/src/Page/Proposal.php
+++ b/api/src/Page/Proposal.php
@@ -314,6 +314,7 @@ class Proposal extends Page
         }
     }
 
+
     # ------------------------------------------------------------------------
     # Get visits for a proposal
     function _get_visits($visit = null, $output = true)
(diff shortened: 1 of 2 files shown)


api/src/Page/Proposal.php
```diff
@@ -393,7 +393,8 @@ function _get_visits($visit = null, $output = true)
         }
 
         if ($this->has_arg('s')) {
-            $where .= " AND s.visit_number LIKE :" . (sizeof($args) + 1);
+            $where .= " AND (s.visit_number LIKE :" . (sizeof($args) + 1) . " OR s.beamlinename LIKE :" . (sizeof($args) + 2) . ")";
+            array_push($args, $this->arg('s'));
             array_push($args, $this->arg('s'));
```
gfrn, line 398: Is this meant to be duplicated?

ndg63276: Yeah, line 396 now has 2 references to $args so you have to add the search term to $args twice. For some reason you can't refer to the same position in $args twice, so you end up repeating lines like this, or looping like on lines 184-5 of the same file.

ndg63276:
This change was reverted, and this PR is replaced by https://github.com/DiamondLightSource/SynchWeb/pull/873

PR #868, opened 2024-11-29
LIMS-1564 - Remove unused dependencies

gfrn:
**JIRA ticket**: [LIMS-1564](https://jira.diamond.ac.uk/browse/LIMS-1564)]

**Summary**:

A lot of dependencies were aliased and had no impact on the application whatsoever. These are being removed, and code where they were referenced were updated to point directly to the node package instead, which makes the code easier to interpret.

Note that this is not exhaustive, and more dead can be removed.

**Changes**:
- Remove unused/aliased dependencies
- Update library imports

**To test**:
- Check Uglymol still works (`/dc/map/id/15928138/aid/107542405`)
- Check the docs page still renders as normal, but smooth scrolling is disabled (run a static file webserver on the `/docs` folder, view the resulting page)
- Check the logistics page still renders as expected (`/logistics`)
- Check visit stats, ensure all charts are rendered as expected (`/stats/visit/cm37235-5`)


ndg63276 (approved):
LGTM

PR #875, opened 2024-12-04
LIMS-1569: Get image viewer rings from dials if possible

ndg63276:
**JIRA ticket**: [LIMS-1569](https://jira.diamond.ac.uk/browse/LIMS-1569)
**JIRA ticket**: [LIMS-507](https://jira.diamond.ac.uk/browse/LIMS-507)

**Summary**:

Instead of Synchweb drawing fixed circles to show resolution or ice rings, ask the dials rest service to do it for you. Fall back to old method if $dials_rest_url not set.

**Changes**:
- Make `$dials_rest_url` available to frontend
- If `$dials_rest_url` is set, ask the backend for resolution or ice rings, else continue drawing them in JS
- Pass request for resolution or ice rings to dials rest service (and cache images)
- Fix bug where `invert` checkbox only worked some of the time

**To test**:
- Set `$dials_rest_url` and `$dials_rest_jwt` to the same as prod
- Go to a recent i03 visit (eg the commissioning visit) so images are still on disk, go to a data collection and click on the diffraction image to enlarge it, check the image appears
- Turn on resolution rings, check 5 rings appear in red. Turn on ice rings, check various rings appear in blue. Turn on both and check both appear.
- Check once they have been loaded, the images appear more quickly as they are cached on the server
- Go to an i19-1 visit and find a data collection with a non-zero value for 2 theta. Repeat the above checks, the rings should now be offset and no longer perfectly circular
- Go to an i23 visit and find a recent data collection. Repeat the above checks, the rings should appear 'squashed' (see LIMS-507)
- Check turning "invert" on and off works as expected
- Unset `$dials_rest_url` and redo the checks on the i03 visit (use a different data collection to avoid caching). Circles should now be drawn by the front end. They will not be accurate for i19-1/i23 visits.

### DIFF
--- a/client/src/js/modules/dc/views/imageviewer.js
+++ b/client/src/js/modules/dc/views/imageviewer.js
@@ -353,6 +353,7 @@ define(['jquery', 'marionette',
             var self = this
             this.ui.canvas.fadeOut(100,function() {
               self.load(n)
+              self.ui.invert.prop('checked', false)
             })
         },
         


gfrn:
Enabling invert before toggling resolution/ice rings negates the value of the invert checkbox

E.g.: If you toggle "invert", then "ice rings", the image is no longer inverted. Unchecking "invert" then inverts the image

Is this expected behaviour?

gfrn (approved):

PR #876, opened 2024-12-09
LIMS-1537: Add more options to reprocessing

ndg63276:
**JIRA ticket**: [LIMS-1537](https://jira.diamond.ac.uk/browse/LIMS-1537)

**Summary**:

Add the d_max option and the absorption level option (for xia2 only) for reprocessing jobs. Also fix a bug in space groups when 'process individually' is ticked.

**Changes**:
- Add a low res input next to the current high res input
- Add a dropdown for absorption level with high/medium/low options, default is left blank
- Add the above changes to standard single crystal reprocessing and mutli crystal reprocessing
- Fix the identifier for the space group box to an input (rather than select) for when 'process individually' is ticked, so that the space group can be populated from the 'Use AP Cell' button, and the requested space group is passed on to the reprocessing job

**To test**:
- Change your config to use the production database otherwise the reprocessing won't work
- Go to a recent data collection (so that images are still on disk) in the commissioning directory (as we are using the prod db) where auto processing has been successful (eg /dc/visit/cm37235-5/id/16115203)
- Click the cog icon for reprocessing, choose a Xia2 pipeline (eg Xia2 Dials). Check there are High Res and Low Res fields, put in something like 2A for High Res and 3A for Low Res
- Click the Xia2 Options button, check there is now a dropdown for Absorption Level, set it to something eg Low
- Set the Start to 1 and the End to 100 to ensure processing is quicker, then press Integrate
- Wait around 10 minutes (maybe start the next test while you wait) for it to complete, then click on the new XIa2 Dials tab, and click on Processing Log. Click the button marked 'xia2 output' and you should get a text file, about the 14th line is the 'Command line' Check it has d_min as your High Res, d_max as your Low Res, and absorption_level=low
- Go back to the reprocessing screen, and this time tick 'Process Individually'. Choose a different pipeline (eg fastdp) and click the 'Space Group / Cell'. Click 'Use AP Cell' and check the space group is populated. Change the space group to something else (eg P1), and press Integrate
- Wait for the job to complete (it may fail as the space group is wrong, don't worry), then click on the Processing Log, then click on 'fast_dp log text'. Check you see a line saying 'Set spacegroup: P 1'
- Go back to the reprocessing screen, and this time click 'Multi Crystal' in the top right. Scroll down to the graphs, drag over a range of images from one of the data collections, then fill in the High and Low Res values as before, use Xia2 3dii as the pipeline, and set the Absorption Level to Medium. Press Integrate.
- As the job starts, you should see the parameters you set at the top of the page. When it completes, you should be able to click on the 'DC' column to go to the relevant data collection and see the processing logs, check your parameters were used by the program as before.

### DIFF
--- a/client/src/js/modules/dc/views/reprocess2.js
+++ b/client/src/js/modules/dc/views/reprocess2.js
@@ -46,7 +46,8 @@ define(['backbone', 'marionette', 'views/dialog',
             ind: 'div.ind',
             pipeline: 'select[name=pipeline]',
             res: 'input[name=res]',
-            sg: 'select[name=SG]'
+            lowres: 'input[name=lowres]',
+            sg: 'input[name=SG]'
         },
 
         events: {
@@ -71,11 +72,16 @@ define(['backbone', 'marionette', 'views/dialog',
 
             'change @ui.pipeline': 'updatePipeline',
             'change @ui.res': 'updateRes',
+            'change @ui.lowres': 'updateLowRes',
         },
 
 
         updateRes: function() {
             this.model.set('RES', this.ui.res.val())
+        },
+
+        updateLowRes: function() {
+            this.model.set('LOWRES', this.ui.lowres.val())
         },
 
         updatePipeline: function() {
@@ -139,6 +145,7 @@ define(['backbone', 'marionette', 'views/dialog',
             this.ui.ind.hide()
             this.$el.find('ul').addClass('half')
             this.$el.find('li input[type="text"]').css('width', '25%')
+            this.$el.find('div input[type="text"]').css('width', '50px')
             this.ui.pipeline.html(this.getOption('pipelines').opts())
             this.model.set('PIPELINE', this.ui.pipeline.val())
         },
@@ -172,7 +179,7 @@ define(['backbone', 'marionette', 'views/dialog',
                 this.ui.be.val(c['CELL_BE']).trigger('change')
                 this.ui.ga.val(c['CELL_GA']).trigger('change')
 
-                this.ui.sg.val(e['SG']).trigger('change')
+                this.ui.sg.val(e.get('SG')).trigger('change')
             }
         },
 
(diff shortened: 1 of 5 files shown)


gfrn:
When going to `/dc/visit/cm37235-5/id/16115203`, the page fails to load because `ENERGY` is not defined.

```
ReferenceError: ENERGY is not defined
    exports webpack://synchweb-webpack/./src/js/templates/dc/dc.html?:66
    Backbone 8
    Underscore 2
    Backbone 5
    onRender webpack://synchweb-webpack/./src/js/modules/dc/datacollections.js?:135```
    
    Does this also happen on your end? This does not happen on `master`.

ndg63276:
> Does this also happen on your end? This does not happen on `master`.

Doesn't happen for me. I assume from the error it is something to do with https://github.com/DiamondLightSource/SynchWeb/pull/853 but not sure why it would appear...

gfrn:
That solved it, but now there's a slightly more annoying issue, pressing "integrate" does nothing, the backend basically ignores the body of `/params` and `/sweep`. I've tried clearing data, switching to Chrome, private mode, nothing helped.

Seems like the body is "discarded" between line 505 and 640 of `api/src/Page.php` (I don't know why we have a custom-built function to merge query parameters and JSON bodies...). Does this happen on your end too?

ndg63276:
> Does this happen on your end too?

Yeah I'm getting that too. Will have a think.

ndg63276:
Ok, found some old Slack chats with John, for me this is caused by some weird cookie or cache thing in the browser. I've cleared my cookies and now it runs, see /dc/visit/mx23694-125/id/16200751. Not sure why that doesn't work for you though...

gfrn:
> Ok, found some old Slack chats with John, for me this is caused by some weird cookie or cache thing in the browser. I've cleared my cookies and now it runs, see /dc/visit/mx23694-125/id/16200751. Not sure why that doesn't work for you though...

It's really odd, sometimes clearing data/cookies works for me, other times it doesn't... I'm going to keep trying, maybe sometime it'll fix itself

gfrn:
image
On the multi crystal reprocessing page, the success message is displayed in red (they're green when dealing with single crystals). Not a huge issue, given SynchWeb's status quo pertaining message colours, but worth keeping in mind, I think!

gfrn (approved):
Tested in `dev-6`, all seems good!

PR #878, opened 2024-12-12
LIMS-1589 - Remove Highcharts

gfrn:
**JIRA ticket**: [LIMS-1589](https://jira.diamond.ac.uk/browse/LIMS-1589)

**Summary**:

To reduce the amount of vulnerabilities, Highcharts was removed, since it was only being used in one page which didn't work properly anyway.

**Changes**:
- Remove Highchart and dewar stats world map

**To test**:
- Go to `shipments/stats`, ensure map is no longer displayed


client/src/js/modules/shipment/views/dewarstats.js
```diff
@@ -141,17 +136,8 @@ define(['marionette',
                 backgrid: { emptyText: 'No dewar stats found' } 
             }))
 
-            this.listenTo(this.countries, 'sync', this.plotMap)
-            this.plotMap()
         },
```
ndg63276, line 140: We still need this call to `plotYears`

```suggestion

        onShow: function() {
            this.listenTo(this.run, 'sync', this.plotYears)
        },
```

ndg63276:
This commit is hard to test, as the query times out. Is it worth removing the "With Data" checkbox, so we can remove the joins to BLSample and DataCollection? Or is that out of scope for this PR?

gfrn:
> This commit is hard to test, as the query times out. Is it worth removing the "With Data" checkbox, so we can remove the joins to BLSample and DataCollection? Or is that out of scope for this PR?

I feel like this is worth its own PR, as fixing this issue might not be so trivial (the page doesn't function currently)

ndg63276 (approved):
LGTM. Will re-open LIMS-995 to fix the other problem.

PR #883, opened 2024-12-16
LIMS-1528: Fix sorting on old summary page

ndg63276:
**JIRA ticket**: [LIMS-1528](https://jira.diamond.ac.uk/browse/LIMS-1528)

**Summary**:

The old data collection summary page is missing a few basic features, eg sorting and filtering

**Changes**:
- Move lots of the logic from js to PHP
- Remove essentially redundant columns, eg date, number of images, exposure time
- Add some more useful columns eg Anom Completeness, CC1/2, CCanom
- Add sorting by clicking on column headers
- Remove weightings from top bar (not well understood), replace with filters

**To test**:
- Go to a visit, then click the "Summary" button to go to eg /dc/summary/visit/mx37593-23
- Check that a list of processed data collections appear
- Check each header works to order the data collections, ascending then descending then reset. Some columns have multiple values, they usually sort by the first value shown, but resolution sorts by the lowest number shown
- Check the filters at the top work to exclude data collections. By default it picks the best pipeline for each data collection, but if you choose a pipeline, it shows only that pipeline
- Check the link to each data collection (magnifying glass icon) works
- Check the download processing archive button (download icon) works
- Check the search bar works on prefix and sample name
- Check completeness values above 95 are green, above 80 are yellow, and below that red


gfrn (approved):

gfrn:
There is an unrelated issue, which I've mentioned before, where you cannot go to the summary page by pasting in the URL. But looks good to me, this issue is unrelated to this PR

PR #891, opened 2025-01-09
LIMS-1466: Disable dispatch form validation when using the shipping service

ndg63276:
**JIRA ticket**: [LIMS-1466](https://jira.diamond.ac.uk/browse/LIMS-1466)

**Summary**:

The dispatch request form has quite strict validation, but little of it is used (or even displayed) if the shipping service will be used. This has led to problems where a hidden field fails validation and the user doesn't know how to fix the problem. We should disable validation if the shipping service is going to be used.

**Changes**:
- Move validation patterns out of model file and into functions within view file
- Tidy some whitespace

**To test**:
- Enable `$use_shipping_service` and `$use_shipping_service_redirect` in config
- Go to a test shipment (eg /shipments/sid/63609) and click "Dispatch"
- Change the country to eg Aruba to show the full form
- Add illegal characters to each field, check that they go red and display an error
- Change the country to United Kingdom, the fields should now be hidden. Check you can Proceed to the shipping service.

### DIFF
--- a/client/src/js/modules/shipment/models/dispatch.js
+++ b/client/src/js/modules/shipment/models/dispatch.js
@@ -12,24 +12,18 @@ define(['backbone'], function(Backbone) {
 
         LOCATION: {
             required: false,
-            pattern: 'wwsdash'
         },
 
         VISIT: {
-            required: function() {
-                return this.visitRequired
-            },
-            pattern: 'visit',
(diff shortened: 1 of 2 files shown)


gfrn:
image
The phone number is not verified when United Kingdom is not selected. Not sure how much of a problem that is?

gfrn (approved):

PR #893, opened 2025-01-13
LIMS-1366: Allow staff direct access to registered dewar page

ndg63276:
**JIRA ticket**: [LIMS-1366](https://jira.diamond.ac.uk/browse/LIMS-1366)

**Summary**:

Staff should be able to access `/dewars/registry/DLS-MX-xxxx` directly, without needing to be on the correct proposal. Also the location history on that page is broken.

**Changes**:
- If a staff member looks at a registered dewar, don't check the proposal
- Search for dewar history by facility code not dewar id
- Fix the total count of registered dewars
- Remove the lab contacts query as it is not used
- Remove the `regdewaradd` page as it is not accessible, dewars are added from the /dewars/registry page

**To test**:
- Log in as a staff member, go to proposal mx34263.
- Go to `/dewars/registry` and check the list of dewars has about 90 pages and the list of proposals has about 1200 pages. Check there are dewars/proposals on the last page of each.
- Add a new dewar eg DLS-MX-0123 with a proposal, check it appears if you search for it
- Add another proposal to the same dewar, check it remains after a refresh
- Search for dewar DLS-MX-0112, and click on it. Check you are shown about 5 associated proposals, about 3 pages of shipments, and about 25 pages of history
- In a new tab, log in as staff and go to `/dewars/registry/DLS-MX-0112` without selecting a proposal, check the view is the same
- Log in as fedid ygr68196 (non-staff), go to proposal mx34263
- Go to `/dewars/registry` and check you are only shown about 4 pages of dewars and not the proposals table or the add dewars section
- Search for dewar DLS-MX-0112, and click on it. Check you are shown about 5 associated proposals, but only about 4 shipments, and only about 4 pages of history
- In a new tab, log in as ygr68196 and go to `/dewars/registry/DLS-MX-0112` without selecting a proposal, check the page doesn't load


gfrn (approved):

gfrn:
Mind if I push a commit that makes the success messages green instead of red for the dewar registry page?

PR #839, opened 2024-09-25
LIMS-128: Add plate view to queue page

ndg63276:
**JIRA ticket**: [LIMS-128](https://jira.diamond.ac.uk/browse/LIMS-128)

**Summary**:

When queuing subsamples in a plate for data collection, it would be useful to show a view of the plate, rather than just a list of (possibly 1000s of) subsamples.

**Changes**:
- Move filtering by point/region/manual/auto to the server
- Display a plate view on the "Prepare For Data Collection" page (/containers/queue/<containerid>)
- Add a search bar to the same page
- Synchronize the search bar with clicking on the plate wells
- Load the inspection image when you click on the wells
- Fix site-wide bug where clearing the search box using the x (Chrome-only?) doesn't clear the search parameters

**To test**:
- Go to a VMXi proposal, eg nt37104, navigate to the /containers page, click a plate that is 'in storage' and has some subsamples, eg /containers/cid/309228
- Click on a few wells with subsamples and assign them a 'score', in the top left of the photo of the well, the colour of the well should change depending on the score you give
- Click "Prepare for Data Collection" to go to eg /containers/queue/309228. Check a view of the plate appears with the same colours as the previous page
- Check the Point/Region/Manual/Auto filters query the server and display only the relevant subsamples
- Click on a well on the plate, check the search bar updates to eg A1d1 and only subsamples matching that appear.
- Click on a well, check the photo of the well changes.
- Edit the search manually, check the results update
- Refresh the page, check the search perseveres (as the page URL has /s/searchquery)
- Delete the search using backspace, check all subsamples are displayed again
- (In Chrome or Edge), do a search, then clear it using the little [x] that appears, check the search updates correctly (this affects all search boxes on the site)

PR #900, opened 2025-01-31
LIMS-128: Add plate view to queue container page

ndg63276:
**JIRA ticket**: [LIMS-128](https://jira.diamond.ac.uk/browse/LIMS-128)

**Summary**:

When queuing samples in a plate for data collection, it would be useful to show a view of the plate, rather than just a list of (possibly 1000s of) samples.
First attempt at this was https://github.com/DiamondLightSource/SynchWeb/pull/839 but had to be reverted.

**Changes**:
- Display a plate view on the "Prepare For Data Collection" page (/containers/queue/)
- Filter the list when you click on a well
- Load the inspection image when you click on a well
- Ensure 'Add all to queue' only adds from the current well (otherwise you get an error as the server queues samples the client isn't showing)

**To test**:
- Go to a proposal with plates (eg nt37104) and then go to a plate (eg /containers/cid/317254)
- Check clicking on a well still loads the inspection image for the well (this page hasn't changed)
- Click "Prepare for Data Collection" near the bottom
- Check a map of the plate appears, and that wells with samples in are outlined (most of them). Check the container name at the top is a link back to the previous page.
- Check that clicking on a well shows a list of samples from that well
- Check the Auto/Manual/Point/Region filters work, alone and in combination with a well
- Check the "Add Current Page to Queue" button adds only those currently displayed
- Check the "Add All to Queue" adds all the pages, but obeys the Auto/Manual/Point/Region filters, and if you have clicked on a well, it obeys that filter also.
- Check the list of "Queued Samples" stays the same even if you change filter on "Available Samples"

### DIFF
--- a/client/src/js/modules/imaging/views/queuecontainer.js
+++ b/client/src/js/modules/imaging/views/queuecontainer.js
@@ -1013,7 +1013,7 @@ define(['marionette',
             let newPos = this.avtypeselector.selectedPos != pos
             this.avtypeselector.selectedPos = null
             this.avtypeselector._filter()
-            _.map(this.subsamples.fullCollection.where({ isSelected: true }), function(ss) {
+            this.subsamples.fullCollection.where({ isSelected: true }).map((ss) => {
                 ss.set({ isSelected: false })
             })
             if (newPos) {
--- a/client/src/js/modules/imaging/views/queuecontainer.js
+++ b/client/src/js/modules/imaging/views/queuecontainer.js
@@ -786,7 +786,6 @@ define(['marionette',
         },
         
         initialize: function() {
-            this._lastSample = null
             this._subsamples_ready = []
 
             this.platetypes = new PlateTypes()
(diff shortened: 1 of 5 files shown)


client/src/js/modules/imaging/views/queuecontainer.js
```diff
@@ -992,12 +999,31 @@ define(['marionette',
...
 
         onShow: function() {
             this.rimg.show(this.image)
         },
         
+        filterByLocation: function(pos) {
+            let newPos = this.avtypeselector.selectedPos != pos
+            this.avtypeselector.selectedPos = null
+            this.avtypeselector._filter()
+            _.map(this.subsamples.fullCollection.where({ isSelected: true }), function(ss) {
```
gfrn, line 1016: Could we use JS's native `map` instead? Just so we reduce reliance on Underscore:

```suggestion
            this.subsamples.fullCollection.where({ isSelected: true }).map((ss) => {
```

gfrn:
When you click a well whilst preparing it for a data collection (say, `/containers/queue/317254`), all other wells turn grey. Is that by design?

ndg63276:
> When you click a well whilst preparing it for a data collection (say, `/containers/queue/317254`), all other wells turn grey. Is that by design?

Not by design as such, but a necessary side-effect of connecting the plate diagram and the collection of available samples. By clicking on a well, you reduce the available samples to just those from that well, so the others are hidden.

The downside I noticed was that if you clicked on a well with no samples, nothing was highlighted at all, which could be confusing. So I've added a commit to always highlight the last clicked well in cyan, even if there were no samples in it.

gfrn (approved):

gfrn:
> > When you click a well whilst preparing it for a data collection (say, `/containers/queue/317254`), all other wells turn grey. Is that by design?
> 
> Not by design as such, but a necessary side-effect of connecting the plate diagram and the collection of available samples. By clicking on a well, you reduce the available samples to just those from that well, so the others are hidden.
> 
> The downside I noticed was that if you clicked on a well with no samples, nothing was highlighted at all, which could be confusing. So I've added a commit to always highlight the last clicked well in cyan, even if there were no samples in it.

That makes sense, this makes things clearer too! Thanks!

PR #901, opened 2025-02-06
LIMS-1389 - Add upload to ccp4 cloud button

ndg63276:
**JIRA ticket**: [LIMS-1389](https://jira.diamond.ac.uk/browse/LIMS-1389)

**Summary**:

We should allow users to directly send processed files (eg mtz, hkl) to CCP4 cloud, without them needing to download/upload. We can add an 'Upload to CCP4 Cloud' button to the 'attached files' dialog.

**Changes**:
- Add a 'CCP4 Cloud' button to the attached files dialog, for both data processing jobs and also data collections themselves (for future work with mmcif files)
- Add a config variable to alter upload path, also allows the button to be removed
- Add a link to help pages on CCP4 cloud (page does not yet exist but Eugene says something will soon)
- Add ability to store ccp4 username/cloudrun id in a cookie
- Return to attached files dialog after successful upload

**To test**:
- Set config variable:
```
$ccp4_cloud_upload_url = 'https://data.cloud.ccp4.ac.uk/api/data/<%=USERNAME%>/<%=FACILITYSHORT%>/<%=VISIT%>_<%=IMAGEPREFIX%>_<%=DATACOLLECTIONNUMBER%>/upload';
```
- Sign up for a CCP4 cloud account at https://cloud.ccp4.ac.uk/
- Go to a visit with processed data eg mx23694-127
- Go to the 'Logs & Files' dialog, check there is a 'CCP4 Cloud' button next to any file of type 'result'
- Click the button, then click 'Upload' with no credentials, check there is a sensible error
- Fill in incorrect credentials, click 'Upload', check there is a sensible error
- Get your cloudrun id from the 'My Account' section of CCP4 cloud, use it to fill in the correct credentials. Tick the box, click 'Upload' and check it seems to run ok, and returns to the attached files dialog.
- Go to CCP4 Cloud, start a new project, and choose the task 'Import from Cloud Storage'. Click 'Select file' and then look in 'My Cloud Storage', the file should be under '/dls/mx23694-127_prefix', depending on the data collection you used.
- There are no files attached to a data collection of type 'params', so insert one (any file will do), or update an existing one, or change attachments.js:56 to 'recip'.
- Go to the data collection and click the little attachments (snowflake) button. Check a 'CCP4 Cloud' button appears next to the relevant attachment, and click it.
- Check your credentials have been remembered, and that the upload works in the same way as before.

### DIFF
--- a/client/src/js/views/cloudupload.js
+++ b/client/src/js/views/cloudupload.js
@@ -1,0 +1,65 @@
+define([
+    'backbone',
+    'views/dialog',
+    'templates/dc/cloudupload.html',
+    'jquery.cookie'
+], function(
+    Backbone,
+    DialogView,
+    template
+) {
+
+    return DialogView.extend({
+        template: template,
+        title: 'Upload to CCP4 Cloud',
+
+        buttons: {
+            'Upload': 'upload',
+            'Cancel': 'closeDialog',
+        },
+        
+        ui: {
+            username: 'input[name=username]',
+            cloudrunid: 'input[name=cloudrunid]',
+            remember: 'input[name=remember]',
+        },
+        
+        upload: function() {
+            if (this.ui.remember.is(':checked')) {
+                let cookieOpts = { expires: 365, path: '/' }
+                $.cookie('ccp4_username', this.ui.username.val(), cookieOpts)
+                $.cookie('ccp4_cloudrunid', this.ui.cloudrunid.val(), cookieOpts)
+            }
+            var self = this
+            Backbone.ajax({
+                url: app.apiurl+'/processing/upload',
+                type: 'POST',
+                data: { id: this.model.get('AUTOPROCPROGRAMATTACHMENTID'), cloudrunid: this.ui.cloudrunid.val(), username: this.ui.username.val() },
+                success: function() {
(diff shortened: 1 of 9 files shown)


client/src/js/views/cloudupload.js
```diff
@@ -0,0 +1,71 @@
...
+        ui: {
+            username: 'input[name=username]',
+            cloudrunid: 'input[name=cloudrunid]',
+            remember: 'input[name=remember]',
+        },
+        
+        upload: function() {
+            if (this.ui.remember.is(':checked')) {
+                let cookieOpts = { expires: 365, path: '/' }
+                $.cookie('ccp4_username', this.ui.username.val(), cookieOpts)
```
gfrn, line 30: Could we use the native cookie management library instead, since jQuery Cookie was removed? I'm happy to do it 

https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/cookies/set
https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/cookies/get

gfrn (approved):

gfrn:
Looks good! We could maybe parse the response in line 72 of `cloudupload.js` so it shows a neater error message, but that is a minor nitpick

PR #903, opened 2025-02-07
LIMS-1612: Replace uniqid with openssl random pseudo bytes

RichB-DLS:
**JIRA ticket:** [LIMS-1612](https://jira.diamond.ac.uk/browse/LIMS-1612)

**Summary**:
Removal of all usage of php's [uniqId](https://www.php.net/manual/en/function.uniqid.php) as its is known to be cryptographically insecure. Replaced with [openssl_pseudo_random_bytes](https://www.php.net/manual/en/function.openssl-random-pseudo-bytes.php) instead.
All uses of uniqId were being passed into md5() so added a common utility fn for that purpose.

**Changes**:
- Added a static utility fn in api/src/Utils.php with a corresponding basic test for params.
- Replaced all uses of uniqId with the new fn

**To test**:
- Go to a dataCollection and attempt to download the processing log. Check the the url query contains the token param.


ndg63276:
Sorry to be a pain, but now that https://github.com/DiamondLightSource/SynchWeb/pull/852 has been merged into master, there is another place we should use the new utility function. Can you merge master into this and then check Shipping.php line 3156 (ish)?

RichB-DLS:
No problem. Will have a look now.

ndg63276 (approved):
Looks good to me.

PR #914, opened 2025-03-13
LIMS-148: Improve staff imaging dashboard

ndg63276:
**JIRA ticket**: [LIMS-148](https://jira.diamond.ac.uk/browse/LIMS-148) and [LIMS-153](https://jira.diamond.ac.uk/browse/LIMS-153)

**Summary**:

Improve the staff imaging dashboard (/admin/imaging)

**Changes**:
- Fix search so it works on more columns
- Fix sorting so it works on all columns
- Display no of queued/completed/available subsamples for both auto and manually generated ones
- Highlight any non-zero value for queued subsamples
- Display when a container is queued
- Hide Puck/Plate/Imager filters if looking at imaging dashboard as redundant

**To test**:
- Log in as someone with "disp_cont" permissions eg rjt57216
- Go to any proposal, then go to /admin/imaging, or click 'Imaging' in the top right
- Check the Containers table loads after a second or two, and all columns display something
- Make a note of the Queued/Completed/Available numbers of a plate with >0 available and is not yet completed, then click on the plate, click on Prepare For Data Collection, then click the + sign next to a subsample. Fill in some numbers for the parameters and click 'Queue Container'
- Go back to /admin/imaging and check the available number has decreased by one and the queued number has increased by one. Check the Completed column now says 'Queued'. Check the queued number is highlighted.
- Check searching works on Name, Barcode and Shipment columns, and also the proposal part of the Visit column
- Check the filters ("Queued", "Completed", "Processing" etc) work
- Check sorting works (slow if on the full list)
- Go to /containers, this table uses the same logic. Check there are extra filters (Plates/Pucks/In Imager). Click the Pucks filter and check the columns change. Check filters, searching and sorting works for this page too.


client/src/js/modules/shipment/views/containers.js
```diff
@@ -141,12 +159,16 @@ define(['marionette',
         },
 
         updateCols: function(selected) {
-            var isPuck = (selected == null || selected == 'puck')
+            var isPuck = !this.imager && (selected == null || selected == 'puck')
```
gfrn, line 162:
```suggestion
            var isPuck = !this.imager && (selected === null || selected === 'puck')
```

Might be useful to use strict equality here - since `null` == `undefined` in non-strict mode

gfrn (approved):

PR #919, opened 2025-03-24
LIMS-1665: Don't allow multiple groups with the same name

ndg63276:
**JIRA ticket**: [LIMS-1665](https://jira.diamond.ac.uk/browse/LIMS-1665)

**Summary**:

Users need to create sample groups when setting up samples for UDC. The 'create container' screen does not indicate when this has succeeded and allows for multiple groups of the same name to be created. These groups are distinct, but indistinguishable for the user. This is causing confusion.

Sample groups should be uniquely named within a proposal and users should receive confirmation when a group is created successfully from the create container page.

**Changes**:
- Add a check that the sample group name is unique within the proposal
- Tidy the `_get_sample_groups_by_sample` function
- Fix bug where error messages were not passed through correctly
- Add success notifications for sample group creation, both from Create Container and Create New Sample Group pages

**To test**:
- Go to a proposal with existing sample groups, eg mx23694
- Go to Sample Groups, then click "Create Sample Group" (or go to /samples/group/new)
- Type a group name that already exists, eg `abcdef`, click Save Sample Group, check a red error message appears and no new sample group is created
- Type a different group name, click Save Sample Group, check a green success message appears and the sample group is created
- Select a shipment, a container and select at least one sample, type a different group name, click Save Sample Group, check 2 green success messages appear and the group is created with samples
- Go to a shipment and click Create Container (or just go to /containers/add/did/78301)
- Click the dropdown for Sample Group in one of the rows of the table. Type a name that already exists (eg `abcdef`), check there is no 'Create New' option. Type a different name, click 'Create New', check a green success message appears.
- Go to a container that already exists. Click the edit button on one of the samples. Type a new sample group name and click 'Create New'. Check a green message appears.
- Click 'Create New' again. Check a red error message appears. Go to the Sample Groups page, check only one group has been created.

### DIFF
--- a/client/src/js/modules/types/mx/samples/sample-table-mixin.js
+++ b/client/src/js/modules/types/mx/samples/sample-table-mixin.js
@@ -292,7 +292,12 @@ export default {
     async createNewSampleGroup(value) {
       this.$emit('update-sample-group-input-disabled', true)
       const sampleGroupModel = new SampleGroup({ NAME: value })
-      await this.$store.dispatch('saveModel', { model: sampleGroupModel })
+      const result = await this.$store.dispatch('saveModel', { model: sampleGroupModel })
+      if (result) {
+        const { NAME } = result.toJSON()
+        let message = 'Created sample group ' + NAME
+        this.$store.commit('notifications/addNotification', { title: 'Success', message: message, level: 'success' })
+      }
       this.$emit('update-sample-group-list')
     },
     canEditRow(location, editingRow) {
--- a/client/src/js/modules/types/mx/samples/sample-table-mixin.js
+++ b/client/src/js/modules/types/mx/samples/sample-table-mixin.js
@@ -300,7 +300,7 @@ export default {
       }
       if (result) {
         const { NAME } = result.toJSON()
-        let message = 'Created sample group ' + NAME
+        const message = 'Created sample group ' + NAME
         this.$store.commit('notifications/addNotification', { title: 'Success', message: message, level: 'success' })
       }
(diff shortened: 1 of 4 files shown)


client/src/js/modules/types/mx/samples/sample-table-mixin.js
```diff
@@ -292,7 +292,17 @@ export default {
...
-      await this.$store.dispatch('saveModel', { model: sampleGroupModel })
+      let result;
+      try {
+        result = await this.$store.dispatch('saveModel', { model: sampleGroupModel })
+      } catch (err) {
+        this.$store.commit('notifications/addNotification', { title: 'Error', message: err.message, level: 'error' })
+      }
+      if (result) {
+        const { NAME } = result.toJSON()
+        let message = 'Created sample group ' + NAME
```
gfrn, line 303:
```suggestion
        const message = 'Created sample group ' + NAME
```

Since this one is a constant (nitpick, shouldn't really have any bearing on the app)

ndg63276:
> When I select a container and at least one sample, it gives me two conflicting error messages - is this expected?

Hmmm... not expected, but I can't reproduce it. That error suggests something is going wrong with cookies maybe? Can you reproduce it, and does it put anything in the console log, or does the API return an error?

ndg63276:
Yeah I've definitely seen this error before, where $this->args becomes empty, so everything fails. Hopefully it will magic itself better tomorrow.

ndg63276:
I've deployed this branch to https://ispyb-dev-1.diamond.ac.uk/ to see if that helps...

PR #921, opened 2025-03-26
LIMS-1648: Only display auto proc program attachment files where deleted column is 0

RichB-DLS:
Ticket 1648 - Only display AutoProcProgramAttachement files where deleted column is 0

**JIRA ticket**: [TICKET-1648](https://jira.diamond.ac.uk/browse/LIMS-1648)
and by extension [TICKET-1685](https://jira.diamond.ac.uk/browse/LIMS-1685)

**Summary**:

Due to future plans for the removal of files from the fs after a certain period, Synchweb should use the BLSession.purgedProcessedData column to assess whether the files for a session have been removed or remain available to download. If a file has been removed from the FS it should present an iCat Link pointing (for the time being) at the project root in iCat. Note that industry proposals will always show 'Deleted' instead.
This iCat link should be amendable in the config.php file.

The exception are any files registered with APPAttachment.filePath containing the string listed in config.php $visit_persist_storage_dir_segment. These are never deleted and should therefore should display their download button even when the visit/BLSession is registered as "purged". 

**Changes**:
- !! CONFIG UPDATE !! Added $visit_persist_storage_dir_segment && $icat_base_url to config.php
- Added the BLSession.purgedProcessedData value to each DC retrieved from api request.
- Updated Attachments dialog to reflect the availability of the file, showing either download buttons, iCat Link or "Removed" as required.
- non-provision of icat_base_url will cause any icatLinks to behave as per industry links, showing "removed" and logging a warning to the console
- Updated the Attachments dialog heading with the text "Note: Removed Attachments (non-industrial) may be reached via the iCat Link."

**To test**:
- Update the values in config.php for $icat_base_url & $visit_persist_storage_dir_segment. 

Reference Example: 
- Open Synchweb to a session on a proposal where BLSESSION.purgedProcessedData === "0"
- Open the Auto or Downstream processing section of a DC & click the "Logs & Files" btn.
- The "Download" & "View" buttons should be displayed for each item.

## Purged Session setup: 
- Open Synchweb to an [industry / non-industry] session where BLSESSION.purgedProcessedData !== "0"
- Open the Auto or Downstream processing section of a DC & click the "Logs & Files" btn.

**Non-industry proposal:**
- [x]  The "Download", "View" or "CCP4 upload" buttons should all be replaced with "iCat" btns linking to $icat_base_url/browse/proposal/PROPOSAL_ID in a new tab.
- [x] Any item containing the persistent storage segment string specified at `$visit_persist_storage_dir_segment` in `config.php` should retain its download/view buttons.

**industry proposal:**
- [x]  The "Download" & "View" buttons should all be replaced with the word "Deleted".
- [x] Any item containing the persistent storage segment string specified at `$visit_persist_storage_dir_segment` in `config.php` should retain its download/view buttons.

---

## Config.php Amendments

iCatURL config (non-industry ONLY):
- [x]  if the `$icat_base_url` is not set  in `config.php` then nothing should be shown.
- [x] A warning will also be logged to the console when opening the dialog.

persistent storage (BOTH):
- [x]  if the `$visit_persist_storage_dir_segment`  is not set  in `config.php` and a session has been purged then all attachments will show removed OR icat link


client/src/js/modules/dc/views/autoprocattachments.js
```diff
@@ -20,23 +22,61 @@ define(['marionette',
...
-                this.$el.append('<a class="vapupload button" href="#"><i class="fa fa-cloud-upload"></i> CCP4 Cloud</a>')
+            else {
+                if (isPurgedSession == true) {
+                    // Append for ANY assets of purged industrial proposals
+                    if (isIndustry == true) { this.$el.html('<div>Deleted</div>') }
+                    else {
+                        // if !industrial Proposal && iCatURL is present in config, this file has been removed from store but may be available in iCat/Archive.
+                        iCatBaseUrl !== ""
+                            ? this.$el.html('<a class="button" href=' + iCatBaseUrl +' target="_blank"><i class="fa fa-external-link"/> iCat</a>') 
+                            : this.$el.html('<div>iCatURL missing</div>')
```
ndg63276, line 70: Maybe just leave a blank space, for a hypothetical other facility using Synchweb but not iCat
```suggestion
                            : this.$el.html('')
```

client/src/js/modules/dc/views/autoprocattachments.js
```diff
@@ -20,23 +22,61 @@ define(['marionette',
...
+                    }
+                }
+                
+                // non purged session. This should be caught in the main if block!
             }
 
             return this
         },
-
+
```
ndg63276, line 79:
```suggestion

```

client/src/js/modules/dc/views/autoprocattachments.js
```diff
@@ -117,6 +183,29 @@ define(['marionette',
...
+         * @returns {URL | undefined} undefined if baseUrl not set OR if proposal begins with "in" | "sw".
+         */
+        getICatProposalRootUrl: function(proposalID) {
+            var iCatBaseURl = app.options.get("icat_base_url");
+            if (!iCatBaseURl) console.warn("'icat_base_url' has not been configured for purged attachments. @see config.php")
+
+            return iCatBaseURl
+                ? new URL("browse/proposal/" + proposalID.toUpperCase() + "/investigation" , iCatBaseURl)
+                : undefined;
+        },
```
ndg63276, line 206:
```suggestion
        getICatProposalRootUrl: function(proposalID) {
            var iCatBaseUrl = app.options.get("icat_base_url");
            if (!iCatBaseUrl) console.warn("'icat_base_url' has not been configured for purged attachments. @see config.php")

            return iCatBaseUrl
                ? new URL("browse/proposal/" + proposalID.toUpperCase() + "/investigation" , iCatBaseUrl)
                : undefined;
        },
```

ndg63276 (commented):
Just a couple of nitpicks, but otherwise looks good, and it all seems to work well!

client/src/js/modules/dc/views/autoprocattachments.js
RichB-DLS: This would leave an empty field if the config is not set up. Might be useful to place something there to flag it to the user eg "config missing" / "Link missing" / a broken-link icon or something?

ndg63276: True, but it's not really a user problem, it's a dev problem. You have a console warning that appears as well.

RichB-DLS: Good point. I''l update and push the changes back up.

PR #925, opened 2025-04-22
LIMS-1694: Allow beamlines to disable user assignment of pucks

ndg63276:
**JIRA ticket**: [LIMS-1694](https://jira.diamond.ac.uk/browse/LIMS-1694)

**Summary**:

Add a config variable to enable or disable user assignment of containers, on a per-beamline basis. 

**Changes**:
- Set up a property in the config.php file to enable/disable
- Pass beamline down through dialogs
- If disabled and not staff, display a warning instead of assigning the container

**To test**:
- Copy the `$only_staff_can_assign` variable from the config_sample.php to your config.php file, but set 'i03' to True
- Log in as a non-staff member (eg ygr68196), go to a visit on i03, click "Assign Containers" to go to eg /assign/visit/mx34263-148
- Try to drag a container from the bottom of the screen to the sample changer near the top, check you get a warning and the container is not assigned
- Try the same on an i04 visit, eg /assign/visit/mx34263-147, check assigning and unassigning works
- Log in as a staff member, try the same tests, check assigning and unassigning works
- Leave one assigned on i03, log in again as non-staff, and try unassigning it, check you get a warning and the container stays assigned

image


client/src/js/modules/assign/views/assign.js
```diff
@@ -91,11 +99,22 @@ define(['marionette', 'backbone', 'views/pages',
...
-                content: 'Are you sure you want to unassign &quot;'+this.model.get('NAME')+'&quot; from sample changer position &quot;'+this.model.get('SAMPLECHANGERLOCATION')+'&quot;?',
-                callback: this.doUnAssign.bind(this, options)
-            })
+            staffOnly = false
+            if (options.bl in app.config.onlyStaffCanAssign) {
+                staffOnly = app.config.onlyStaffCanAssign[options.bl]
+            }
+            console.log(staffOnly)
+            console.log(app.staff)
+            console.log(options)
```
gfrn, line 108: Should we remove these, if they were used for debugging purposes only?

gfrn:
Is there a backend check for this? Maybe that would make things safer

PR #929, opened 2025-04-29
LIMS-1447: Don't allow shipping of red samples

ndg63276:
**JIRA ticket**: [LIMS-1447](https://jira.diamond.ac.uk/browse/LIMS-1447)

**Summary**:

Users should not be allowed to ship red samples/shipments through Synchweb. Also, red shipments should already have a red session scheduled (and the same for yellow).

**Changes**:
- Disable the "Create Air Waybill" button if the shipment is rated red.
- If a shipment is being created and is rated as not green, only show suitable sessions, also show a message if no sessions are found

**To test**:
- Find a proposal with an upcoming yellow session, the calendar is good for this
- Go to that proposal and start creating a shipment
- Set the shipment as Green, you should be able to select any option for the "Scheduling" question. Choose "I have a session already scheduled", and check a dropdown appears showing all current and future visits
- Change the shipment to Yellow, check the dropdown now only shows the upcoming yellow visit
- Change the shipment to Red, check the dropdown disappears and a message says "(no suitable sessions found)"
- Change the shipment back to green and finish making the shipment, no containers are necessary.
- View the shipment, check the "Create Air Waybill" button is enabled
- Change= the shipment to Red, refresh the page, check the "Create Air Waybill" button is now disabled, with a message at the top.

### DIFF
--- a/api/src/Page/Proposal.php
+++ b/api/src/Page/Proposal.php
@@ -58,6 +58,7 @@ class Proposal extends Page
         'BEAMLINESETUPID' => '\d+',
         'BEAMCALENDARID' => '\d+',
         'VISITNUMBER' => '\d+',
+        'RISKRATING' => '\w+',
 
         // visit has person
         'SHPKEY' => '\d+\-\d+',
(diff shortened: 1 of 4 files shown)


gfrn:
I was able to create an airway bill for a red shipment

image

To reproduce:
- Create a green shipment
- Change the safety level to red
- Click "create airway bill" without refreshing the page
- Fill in the form
- Create the airway bill

Maybe we also need a backend check to make sure that this can't happen?

ndg63276:
> I was able to create an airway bill for a red shipment

I've moved some of the logic out of the template file and into the JS file, so the page can now change as soon as the model changes. So hopefully that is no longer possible.

gfrn:
Should there be a check in place in the backend too?

I can still create a shipment by sending POST requests or manually navigating to `https://local-oidc-test.diamond.ac.uk:8082/shipments/awb/sid/71136`

ndg63276:
It's hard to bloick on the backend as it may go to the shipping service, impossible to block everything anyway, so I've added a check on `shipments/awb/sid/<shippingId>` to prevent accidental use.

I've also put in some code to encourage users to declare a safety level if they haven't already (eg CSV uploads), by hiding some buttons (with a message to tell them what to do).

PR #934, opened 2025-05-08
[LIMS-1636] Make LDAP server ID field configurable

gfrn:
**JIRA ticket**: [LIMS-1636](https://jira.diamond.ac.uk/browse/LIMS-1636)

**Summary**:

This PR allows users to configure which field is used to get the user ID from in LDAP, to make it easier to switch between LDAP providers.

**Changes**:
- Add `ldap_id_field` configuration key

**To test**:
- Go to proposal `mx23694`, navigate to the shipments page, add a new shipment
- Click `transfer` on the dewar that was just created, check if the LC email is pre-populated
- Click `dispatch` for the same dewar, ensure the same is still true
- If you have dispatch/transfer emails set up, and you're set as one of the recipients, then you should be sent an email. If you are dumping emails to `email.txt`, then these emails should appear in there
- OPTIONAL: there is an unused function named `_get_name` to get names from LDAP, you can modify the code to call it and ensure it still works

### DIFF
--- a/api/config_sample.php
+++ b/api/config_sample.php
@@ -40,6 +40,8 @@
     # CAS CA Cert (for SSO)
     $cacert = '/etc/certs/ca-bundle.crt';
 
+    # Field to get user ID from in LDAP
+    $ldap_id_field = "cn";
     # ldap server, used for lookup and authentication (if using, set to null if not)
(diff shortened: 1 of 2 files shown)


ndg63276 (approved):
LGTM

PR #943, opened 2025-06-03
LIMS-1752: Create view for Ligand Fit pipeline

ndg63276:
**JIRA ticket**: [LIMS-1752](https://jira.diamond.ac.uk/browse/LIMS-1752)

**Summary**:

The new ligand fit pipeline is now running, and generating Molstar model files. We should have a proper view of them.

**Changes**:
- Create a new downstream type for LigandFit
- Make a table with basic stats (similar to shelxt)
- Show the SMILES as a png
- Pass down the AutoProcProgramAttachmentId for the Molstar map HTML page, make clicking on the SMILES image take you to the model 
- Hijack the "Map / Model Viewer" button to take you to the Molstar model as well

**To test**:
- Find a data collection with a successful ligand fit run (eg /dc/visit/nt37477-28/id/15841906. there are several runs but look at the last one)
- Click on Downstream Processing and then LigandFit
- Check it shows what triggered it eg "LigandFit triggered by Dimple"
- Check it shows the Smiles code and the fitting CC value
- Check it shows the molecule as a PNG
- Check clicking on the PNG or the "Map / Model Viewer" button takes you to a new model viewer, and that you can pan/tilt/zoom in the new viewer.

### DIFF
--- a/api/src/Downstream/Type/LigandFit.php
+++ b/api/src/Downstream/Type/LigandFit.php
@@ -1,0 +1,73 @@
+<?php
+
+namespace SynchWeb\Downstream\Type;
+
+use SynchWeb\Downstream\DownstreamPlugin;
+use SynchWeb\Downstream\DownstreamResult;
+
+class LigandFit extends DownstreamPlugin {
+    var $has_images = true;
+    var $friendlyname = 'LigandFit';
+    var $has_mapmodel = array(1, 0);
+
+    function _get_ligandfit_results_json() {
+        $appid = array($this->autoprocprogramid);
+        $filepath = $this->db->pq(
+            "SELECT app.filePath, app.filename from autoprocprogramattachment app where autoprocprogramid = :1 and filename like '%.json' ",
+            $appid
+        );
+        return $filepath;
+    }
+
+    function _get_ligandfit_results_png() {
+        $appid = array($this->autoprocprogramid);
+        $filepath = $this->db->pq(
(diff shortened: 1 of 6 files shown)


api/src/Downstream/Type/LigandFit.php
```diff
@@ -0,0 +1,73 @@
...
+        $dat['SOLUTIONS'] = json_decode($json_data);
+        $dat['PARENTAUTOPROCPROGRAM'] = $this->process['PROCESSINGCOMMENTS'];
+
+        $results = new DownstreamResult($this);
+        $results->data = $dat;
+
+        return $results;
+    }
+
+    function images($n = 0) {
```
gfrn, line 65: Is `$n` used in this function?

api/src/Downstream/Type/LigandFit.php
```diff
@@ -0,0 +1,73 @@
...
+
+namespace SynchWeb\Downstream\Type;
+
+use SynchWeb\Downstream\DownstreamPlugin;
+use SynchWeb\Downstream\DownstreamResult;
+
+class LigandFit extends DownstreamPlugin {
+    var $has_images = true;
+    var $friendlyname = 'LigandFit';
+    var $has_mapmodel = array(1, 0);
```
gfrn, line 11: Is this var unused?

api/src/Downstream/Type/LigandFit.php
ndg63276: No, but it has to be declared to be compatible with the definition in DownstreamProcessing.php. I guess `$n = null` might be better?

gfrn (approved):

PR #945, opened 2025-06-06
LIMS-1739: Add button to say shipment is ready for scheduling

ndg63276:
**JIRA ticket**: [LIMS-1739](https://jira.diamond.ac.uk/browse/LIMS-1739)

**Summary**:

If a user has selected "I would like a session to be scheduled" when creating a shipment, it is hard to know when they have created all the dewars and pucks within the shipment. We should have a button to say "My shipment is ready for scheduling". When the user has pressed this button, we should disable creating new dewars and pucks.

**Changes**:
- Scheduling status is stored as JSON in the Shipping.extra column, under the 'DYNAMIC' key
- Get scheduling status when getting dewars, so can hide the 'Add Container' button
- Add new button for users to click when a shipment has all the dewars and pucks created
  - Only show the new button if the scheduling status is 'I would like a session to be scheduled'
  - Dont show the new button for industrial proposals
  - Show a confirmation dialog when clicked
  - Once confirmed, remove the scheduling status dropdown, the Add Dewar button, and the Add Container button
- Add a remove function to editable for possible re-use
- Move definition of industrial proposal codes to the config

**To test**:
- Add this to the config
```
$industrial_prop_codes = array('sw', 'in', 'ic');
```
- Log in as a non-staff member, eg eyg28471
- Go to an mx proposal and create a shipment, choosing 'I would like a session to be scheduled', and click 'Add Shipment'
- When you are taken to the shipment page, check the 'Scheduling' field says 'I would like a session to be scheduled', and there is a new button alongside it that says 'Shipment is ready for scheduling'
- Check the Scheduling field is an editable dropdown, and the button disappears if another value is chosen from the list. Change it back to 'I would like a session to be scheduled'
- Add a container to the dewar, then go back to the shipment page
- Click the 'Shipment is ready for scheduling' button, check a confirmation dialog appears. Check Cancel just closes the dialog.
- Click the button again and click 'Ok'. Check the Scheduling field now says 'I would like a session to be scheduled - my shipment is ready for scheduling'. Check it is now non-editable, and the button has disappeared.
- Check the 'Add Dewar' and 'Add Container' buttons have disappeared.
- Log in as a staff member and view the shipment. Check you can edit the dropdown to set the shipment back to 'I would like a session to be scheduled'.
- Go to an industrial proposal, eg ic39117. Make a shipment, choosing 'I would like a session to be scheduled', and click 'Add Shipment'
- When you are taken to the shipment page, check the 'Scheduling' field says 'I would like a session to be scheduled', but the new button is not displayed

### DIFF
--- a/client/src/js/modules/shipment/views/shipment.js
+++ b/client/src/js/modules/shipment/views/shipment.js
@@ -54,11 +54,14 @@ define(['marionette',
             'click a.send': 'sendShipment',
             'click a.pdf': utils.signHandler,
             'click a.cancel_pickup': 'cancelPickup',
+            'click a.ready': 'markAsReady',
         },
 
         ui: {
             add_dewar: '#add_dewar',
             longwavelength: '.longwavelength',
+            ready: '.ready',
+            dynamic: '.DYNAMIC',
         },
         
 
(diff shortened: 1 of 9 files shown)


client/src/js/modules/shipment/views/shipment.js
```diff
@@ -279,10 +308,22 @@ define(['marionette',
                     this.$el.find(".remoteform").show()
                 }
             }
+            if (dynamic === 'Ready' || (app.proposal && app.proposal.get('ACTIVE') != '1')) {
```
gfrn, line 311: Is it worth keeping the proposal active comparison as is? With how JS does these comparisons, 1 == 1, but 1 == "1", "1" == true, and so on

client/src/js/modules/shipment/views/shipment.js
```diff
@@ -291,46 +332,49 @@ define(['marionette',
...
                 'UDC': 'I am sending pucks for Unattended Data Collection',
                 'Imaging': 'I am sending plates for imaging',
                 'Yes': 'I would like a session to be scheduled',
                 'Other': 'Something else',
-            }})
-            industrial_codes = ['in', 'sw']
-            industrial_visit = industrial_codes.includes(app.prop.slice(0,2))
-            if (!industrial_visit) {
+            }
+            if (app.staff || this.model.get('DYNAMIC') != 'Ready') {
```
gfrn, line 358:
```suggestion
            if (app.staff || this.model.get('DYNAMIC') !== 'Ready') {
```

PR #948, opened 2025-06-20
LIMS-1501: Allow re-running downstream processing

ndg63276:
**JIRA ticket**: [LIMS-1501](https://jira.diamond.ac.uk/browse/LIMS-1501)

**Summary**:

Add the ability to re-trigger downstream processing jobs.

**Changes**:
- Add a button to Auto Processing jobs to bring up a new re-processing dialog
- Display available downstream jobs (based on what @isikhar advises is available)
- Disable jobs if the relevant PDB/sequence/anomalous scatterer are not defined
- Submit button sends a message over rabbitmq to zocalo to enqueue the job

**To test**:
- Edit your config to use the production database, as Zocalo will place results there
- Find a data collection with sequence, PDB file, and anomalous scatterer (eg /dc/visit/mx23694-139/id/18302905), and open up the 'Auto Processing' tab.
- Check that each tab has a new button 'Downstream Processing'
- Check that it opens a new dialog showing the data collection and a list of pipelines
- Check that the following pipelines are supported (others should show a message to say unsupported)
  - fast_dp - dimple, fast_ep
  - xia2 dials, xia2 3dii, multi-xia2, xia2 multiplex - dimple, MrBUMP, fast_ep, big_ep
  - autoPROC, autoPROC+staraniso - fast_ep, big_ep
- Submit a job, check no errors appear in the dev console.
- Check a new row appears in the ProcessingJob table, with a recipe of eg postprocessing-dimple
- Go to a data collection with no sequence, PDB or anomalous scatterer (eg /dc/visit/mx34263-158/dcg/15347173), check each pipeline is disabled, with a reason stated
- Switch config back to restore db if needed


client/src/js/modules/dc/views/downstream2.js
gfrn, line 1: Can we use a different filename, so that the purpose of this file is a bit more obvious?

ndg63276:
> Seems like the row isn't created on my end (at least if there's no RMQ vhost configured)... Is that expected?

Sorry, yes, that isn't very clear, The row is created by Zocalo, not by Synchweb, so if there is no RMQ vhost, you won't see one.

gfrn (approved):

client/src/js/modules/dc/views/downstream2.js
ndg63276: Yeah the name was to mirror the reprocess2.js file in the same directory, but you're right, better to use something that actually describes it. Have renamed to downstreamreprocess.js.

PR #959, opened 2025-07-15
LIMS-1796: Display filepath for attachments

ndg63276:
**JIRA ticket**: [LIMS-1796](https://jira.diamond.ac.uk/browse/LIMS-1796)

**Summary**:

Where autoprocessing jobs have run is a bit of a mystery within Synchweb. But one thing we can do simply is to display the filepath to any attachments.

**Changes**:
- Add a file path column to the data collection Attachments view, auto processing attachments view and the downstream processing attachments view
- Truncate the path to prevent it being too wide, but display the full path in a tooltip, and add a button to copy to the clipboard

**To test**:
- Deploy to a https server, otherwise clipboard integration doesn't work. At the time of writing this branch is on ispyb-dev-1
- Go to a data collection with attachments, auto processing attachments and downstream processing attachments (eg /dc/visit/mx23694-139/id/18302860)
- Click the little Attachments icon (next to the reprocessing cog), check that each attachment shows the truncated path, with the full path as a tooltip, with a button to copy it to the clipboard. Check the attachments can still be viewed or downloaded.
- Go to an auto processing job and click "Logs & Files", do the same checks
- Go to a downstream processing job and click "Logs & Files", do the same checks.

### DIFF
--- a/client/src/js/modules/dc/views/attachments.js
+++ b/client/src/js/modules/dc/views/attachments.js
@@ -17,7 +17,7 @@ define(['marionette',
             var displayedPath = filePath.split('/').slice(0,4).join('/')+'/...'
 
             this.$el.html(`
-                <span>${displayedPath}</span>
+                <span title="${filePath}">${displayedPath}</span>
                 <button class="copy-path button" title="Copy full path to clipboard">
                     <i class="fa fa-clipboard"/>
                 </button>
--- a/client/src/js/modules/dc/views/attachments.js
+++ b/client/src/js/modules/dc/views/attachments.js
@@ -14,7 +14,7 @@ define(['marionette',
 
         render: function() {
             var filePath = this.model.get('FILEPATH')
-            var displayedPath = filePath.split('/').slice(0,4).join('/')+'/...'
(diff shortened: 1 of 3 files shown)


client/src/js/modules/dc/views/attachments.js
```diff
@@ -7,6 +7,44 @@ define(['marionette',
...
     function(Marionette, Backgrid, TableView, attachments, LogView, CloudUploadView, utils) {
 
+    var PathCell = Backgrid.Cell.extend({
+        events: {
+            'click .copy-path': 'copyPathToClipboard'
+        },
+
+        render: function() {
+            var filePath = this.model.get('FILEPATH')
+            var displayedPath = filePath.split('/').slice(0,4).join('/')+'/...'
```
gfrn, line 17: Nitpick, but if the full path is less than 4 levels deep, this would display ellipsis regardless, it seems

Does that seem like a problem?

ndg63276: Probably not a problem, filepaths are always going to start /dls/beamline/data/...,, but you're correct and it's an easy fix so I've added something to cope with that.

gfrn (approved):

PR #964, opened 2025-07-25
LIMS-1839: Fix fast_dp radiation damage plots

ndg63276:
**JIRA ticket**: [LIMS-1839](https://jira.diamond.ac.uk/browse/LIMS-1839)

**Summary**:

Fast dp radiation damage plots stopped working when some fast_dp files were moved to a different folder.

**Changes**:
- Search specifically for the xdsstat.log file, not just any fast_dp file

**To test**:
- Go to a recent data collection with successful fast_dp (eg /dc/visit/cm40607-3/id/19021858)
- Click on Auto Processing, choose the fast_dp tab, then click Radiation Damage. Check a plot appears.
- Repeat for an older data collection before the files were moved (eg /dc/visit/cm40607-1/id/17183638)

### DIFF
--- a/api/src/Page/DC.php
+++ b/api/src/Page/DC.php
@@ -1621,15 +1621,15 @@ class DC extends Page
                 INNER JOIN autoproc ap ON aps.autoprocid = ap.autoprocid 
                 INNER JOIN autoprocprogram app ON api.autoprocprogramid = app.autoprocprogramid 
                 INNER JOIN autoprocprogramattachment appa ON appa.autoprocprogramid = app.autoprocprogramid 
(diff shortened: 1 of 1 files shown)


gfrn:
Does this warrant a configuration option, or is it not likely to change in the future? (judging by the Slack thread, I don't think it'll change much, but just in case)

ndg63276:
> Does this warrant a configuration option, or is it not likely to change in the future? (judging by the Slack thread, I don't think it'll change much, but just in case)

What would you put the config option on, the filename? I don't think it will change.

gfrn:
> > Does this warrant a configuration option, or is it not likely to change in the future? (judging by the Slack thread, I don't think it'll change much, but just in case)
> 
> What would you put the config option on, the filename? I don't think it will change.

Would be, but since it won't change often, I guess that's not needed

PR #979, opened 2025-09-08
LIMS-1876: Fix URL parameters on fault report list

ndg63276:
**JIRA ticket**: [LIMS-1876](https://jira.diamond.ac.uk/browse/LIMS-1876)

**Summary**:

When viewing fault reports, if you apply a filter, you get a URL like https://ispyb.diamond.ac.uk/faults/bl/i03/sys/null/com/null/sub/null, but refreshing that URL gives an error.

**Changes**:
- Add search parameter to URL
- Fix passing of parameters from router to page
- Force search parameter to end of the URL if a filter is selected
- Don't put null's in the URL
- Update the fault list if filters are selected on first load of the page

**To test**:
- Go to the fault report list page /faults
- Select a beamline (eg i03) from the dropdown, check the list filters correctly, and the list of systems populates
- Check the URL becomes /faults/bl/i03, without any null values
- Check that refreshing the page keeps the beamline field populated and the list filtered
- Repeat the tests for Systems, Components and Subcomponents.
- Check if you click into a fault report, and then click Back, the page loads correctly
- Check the search bar behaves as expected, even if you change the filters after searching

### DIFF
--- a/client/src/js/modules/fault/views/filters.js
+++ b/client/src/js/modules/fault/views/filters.js
@@ -45,7 +45,7 @@ define(['marionette',
 
             _.each(frags, function(v, f) {
                 url = url.replace(new RegExp('\\/'+f+'\\/\\w+'), '')
-                if (this.ui[v].val() && this.ui[v].val() != 0) url += '/'+f+'/'+this.ui[v].val()
+                if (this.ui[v].val()) url += '/'+f+'/'+this.ui[v].val()
             }, this)
             
             url += searchFrag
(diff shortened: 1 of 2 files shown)


client/src/js/modules/fault/views/filters.js
```diff
@@ -27,19 +27,28 @@ define(['marionette',
...
+            const searchMatch = url.match(/\/s\/.*$/)
+            if (searchMatch) {
+                searchFrag = searchMatch[0]
+                url = url.replace(searchFrag, '')
+            }
+
             _.each(frags, function(v, f) {
                 url = url.replace(new RegExp('\\/'+f+'\\/\\w+'), '')
-                if (this.ui[v].val() != 0) url += '/'+f+'/'+this.ui[v].val()
+                if (this.ui[v].val() && this.ui[v].val() != 0) url += '/'+f+'/'+this.ui[v].val()
```
gfrn, line 48:
```suggestion
                if (!!this.ui[v].val()) url += '/'+f+'/'+this.ui[v].val()
```

```suggestion
                if (this.ui[v].val()) url += '/'+f+'/'+this.ui[v].val()
```

Is there any reason why we need the "!= 0"? As 0 is already falsy in JS

As an added measure of precaution, we can even cast the val to a bool with `!!` before evaluating (that means null, undefined, 0, false will all evaluate to false)

PR #1054, opened 2026-03-23
LIMS-2126: Add new strategy pipeline and tidy code

ndg63276:
**JIRA ticket**: [LIMS-2126](https://jira.diamond.ac.uk/browse/LIMS-2126)

**Summary**:

Phil is adding a new strategy pipeline ("UDC Strategy"), we should display the results appropriately. Also the strategy fetching code can be optimised as lots is not used.

**Changes**:
- Fix the getting of the pipeline description, as in MariaDB `if("string")` returns false
- Order screening results by the order they are put into the db
- Use consistent notation to control the number format of the fields
- Remove various fields that were set but never used, and the functions used to set them
- On the front end, only display the unit cell if some of it is set
- Only show chi/phi/kappa/ranking res columns if the are set

**To test**:
- Set in the config
```
$strat_align = array('XOalign', 'dials.align_crystal', 'xia2-dials+dials.align_crystal', 'UDC strategy');
```
- Go to /dc/visit/mx23694-154/id/21392025, check a UDC strategy appears in the strategies bar, and the omega and chi angles are shown. (The relative transmission is set wrong in the db so the absolute transmission is displayed as 0)
- Check the alignment strategies show the chi and phi angles and the unit cell (but no omega or transmission etc)
- Go to a data collection with mosflm strategy results (eg /dc/visit/cm44138-1/id/21135234), check Mosflm shows both the unit cell and the data collection parameters
- Find an old data collection with EDNA strategy results (eg /dc/visit/cm19645-4/id/2983769), check EDNA shows both the unit cell and the data collection parameters)

### DIFF
--- a/api/src/Page/DC.php
+++ b/api/src/Page/DC.php
@@ -1286,7 +1286,7 @@ class DC extends Page
                 dc.imageprefix as imp, dc.comments as dcc, dc.blsampleid as sid,
                 sl.spacegroup as sg, sl.unitcell_a as a, sl.unitcell_b as b, sl.unitcell_c as c, sl.unitcell_alpha as al, sl.unitcell_beta as be, sl.unitcell_gamma as ga,
                 det.numberofpixelsx, det.detectorpixelsizehorizontal,
-                CONCAT(IF(sssw.comments, sssw.comments, IF(ssw.comments, ssw.comments, s.shortcomments)), ' Wedge', IFNULL(ssw.wedgenumber, '')) as com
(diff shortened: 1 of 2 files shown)


gfrn:
image
Seems like the chi angles are missing for UDC strategies, has the database been overwritten or am I missing something?

ndg63276:
Hmmm... that's not right. Mine still shows the chi angles...
image
Has the front end been re-built?

gfrn:
Although I ran it thrice, it was probably cached and didn't get picked up, it worked today for some reason... I'll hit approve on it

gfrn (approved):

PR #1062, opened 2026-04-15
LIMS-1949: Use parentAutoProcProgramId field

ndg63276:
**JIRA ticket**: [LIMS-1949](https://jira.diamond.ac.uk/browse/LIMS-1949)

**Summary**:

It is useful to be able to see a downstream processing job's upstream parent, but at the moment we use the "scaling_id" parameter in the ProcessingJobParameter table, which is a bit flaky. We now have an `AutoProcProgram.parentAutoProcProgramId` field, which is being populated from about 1st April 2026.

**Changes**:
- Re-route all calls to `_lookup_autoproc` to a new `_lookup_parent_autoproc` function, which uses the parentAutoProcProgramId field
- If no parent job is found, fall back to the old method, to avoid all pre-April 2026 jobs being broken
- Use a config variable to limit falling back to before a certain job, as from now on the parentAutoProcProgramId field should be populated
- Simplify the old method a little, Big EP was using the program_id instead of the scaling_id, but this seems not to be necessary

**To test**:
- Go to a recent data collection with lots of downstream processing jobs, eg /dc/visit/nt44198-3/id/22397399
- On each downstream job, click the "View Upstream" button, check it opens the Automatic Processing bar if needed and switches to the correct tab
- Go to an older data collection eg /dc/visit/nt44198-3/id/21167691, repeat the check
- Add the config variable with a low number eg
```
$max_appid_without_parent = 123;
```
- Repeat the check on the older data collection, it should stop working as you have now disallowed falling back to the old method
- Now set it to 119440000, and repeat the check on the older data collection, it should work again.

### DIFF
--- a/api/config_sample.php
+++ b/api/config_sample.php
@@ -439,6 +439,9 @@
         ),
     );
 
+    # Set the first AutoProcProgram row with parentAutoProcProgramId not null
+    #$max_appid_without_parent = 123456789;
+
 
     # Add a button to upload file to CCP4 cloud
     #$ccp4_cloud_upload_url = 'https://data.cloud.ccp4.ac.uk/api/data/<%=USERNAME%>/<%=FACILITYNAME%>/<%=IMAGEPREFIX%>_<%=DATACOLLECTIONNUMBER%>/upload';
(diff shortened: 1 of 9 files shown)


gfrn:
When I set `$max_appid_without_parent` to 123, it does stop working, but silently, it still opens the accordion and sets the tab to `autoPROC`, will this be a problem? I assume in prod, we'll just set to a realistic value anyway

ndg63276:
Yeah it is a bit of a weird test, when we put a realistic number it should behave more normally.

gfrn (approved):

PR #1069, opened 2026-05-08
LIMS-1683: Show buttons for next/previous container on plate view

ndg63276:
**JIRA ticket**: [LIMS-1683](https://jira.diamond.ac.uk/browse/LIMS-1683)

**Summary**:

https://github.com/DiamondLightSource/SynchWeb/pull/912 added buttons to move to the next/previous container in the puck view, this PR is to add the same buttons to the plate views.

**Changes**:
- Move scoped CSS for the header row and the flat buttons into the general use area
- Add prev/next buttons to the view both of plates that have been imaged and those that haven't
- Add code to get sibling containers, work out the current container number, and which container ids are next/previous, to add titles to the buttons, and disable when necessary
- Always do the container sorting in the backend for consistency, and use container id to distinguish containers with the same name
- Remove unused function `updateContainerSampleGroupsData` and related functions `fetchDewars` and `fetchContainers` and related variable `containersCollection`.

**To test**:
- Go to a puck with siblings (eg /containers/cid/367188), check the Next and Prev Container buttons still appear, look the same as on prod, have titles when you hover, and disable correctly when on the first / last puck of the dewar.
- Go to a plate that has been imaged (eg /containers/cid/373688), check the Next and Prev Container buttons appear and behave the same as for pucks
- Go to (or create) a plate that has not been imaged (eg /containers/cid/373685), check the Next and Prev Container buttons appear and behave the same as for pucks
- Add a puck and a plate to a dewar with an already imaged plate, check you can move between the 3 kinds as expected


client/src/js/modules/types/mx/shipment/views/mx-container-view.vue
```diff
@@ -592,43 +589,12 @@ export default {
...
-          c.DEWARID === this.container.DEWARID
-        );
-
-      } else {
-        result = new Containers();
-        result.dewarID = this.container.DEWARID;
-        await result.fetch();
-      }
-
+      result = new Containers(null, { state: { pageSize: 9999 } });
```
gfrn: Nitpick, but could result be initialised like

`let result = new Containers(null, { state: { pageSize: 9999 } });`?

client/src/css/partials/_utility.scss
```diff
@@ -316,3 +316,81 @@ ul.ui-autocomplete {
...
+    &.warning {
+        @apply tw-bg-yellow-300 tw-text-black;
+        &:hover, &:focus {
+            @apply tw-bg-yellow-500;
+        }
+    }
+    &.danger {
+        @apply tw-bg-red-500 tw-text-white;
+        &:hover, &:focus {
+            @apply tw-bg-red-700
```
gfrn:
```suggestion
            @apply tw-bg-red-700;
```

client/src/js/modules/shipment/views/containerplate.js
```diff
@@ -746,13 +756,61 @@ define(['marionette',
...
 
         updatePipelines: function() {
             this.ui.pipeline.html('<option value="">All pipelines</option>'+this.processing_pipelines.opts())
         },
 
+        updateSiblingContainers: function() {
+            let foundCurrent = false;
+            this.siblingContainers.each((container, index) => {
+                let idx = index+1
+                if (foundCurrent && !this.nextContainer) {
```
gfrn, line 775: Should this come after `foundCurrent` is set? (line 779) 

As it currently is, it might just not reflect the actual condition because it is evaluated before the `containerId` check is executed

ndg63276: No, this is deliberate.  You are looping through all the containers (in order), so you need to have already found the currently-displayed one, (and not yet set the 'next' one), in order to set this container as the 'next' one.

gfrn: Yeah, that makes sense, I'll approve it now, thanks for having a look!

gfrn (approved):

PR #1070, opened 2026-05-15
LIMS-2181: Improve speed of queueing and editing 1000s of samples

ndg63276:
**JIRA ticket**: [LIMS-2181](https://jira.diamond.ac.uk/browse/LIMS-2181)

**Summary**:

As we now automatically detect crystals in plates, there can be 1000s of them, that all need queueing, and also parameters adding to. This is currently very slow as it loops through each subsample individually. This change is to make the "Add All to Queue" and "Apply to All" buttons use far fewer database calls.

**Changes**:
- Make the "Add all to Queue" and "Unqueue All" buttons insert or delete 500 rows at a time, using "?" as a placeholder instead of ":1", ":2" etc (seems to work)
- Make the "Apply to All" button call a new endpoint, which first retrieves a list of queued subsamples in the container, then applies the preset to them all in chunks of 500.
- Remove manual validation of all queued samples when queueing the container, as this is very slow (about 200 samples / sec)
- Make queueing and unqueueing the container silent operations, as there is no need to re-render the samples.

**To test**:
- Find a container with 1000s of subsamples (that has not yet been disposed) eg /containers/cid/377216
- Click "Prepare For Data Collection" at the bottom to go to eg /containers/queue/377216
- Assuming none are currently queued (ie the queued samples list is empty), click "Add All to Queue". Check all 7626 (or however many) samples are queued (in about 10s)
- Choose one of the presets that starts with "Omega Scan" from the dropdown, then click "Apply to All". Check this returns in about 3s and each of the parameters updates correctly. (If there are any "regions" in the list they won't be edited as this is a point preset)
- Manually edit the parameters of a queued sample so they are invalid, and then click "Queue Container" at the bottom. Check you are told one is invalid and are taken to the invalid tab to correct it
- Fix the parameters, queue the container, check it works promptly. Check an Unqueue button appears at the top.
- Check the Unqueue button works promptly, and the Unqueue button disappears.
- Click "Unqueue All" and check all the samples are removed from the queue.

### DIFF
--- a/api/src/Page/Sample.php
+++ b/api/src/Page/Sample.php
@@ -177,6 +177,7 @@ class Sample extends Page
         array('/sub/:ssid', 'get', '_get_sub_sample'),
         array('/sub/:ssid', 'patch', '_update_sub_sample'),
         array('/sub/:ssid', 'put', '_update_sub_sample_full'),
+        array('/sub/cid/:cid', 'post', '_update_container_sub_samples'),
         array('/sub', 'post', '_add_sub_sample'),
         array('/sub/:ssid', 'delete', '_delete_sub_sample'),
         array('/sub/queue/cid/:cid', 'post', '_queue_all_sub_samples'),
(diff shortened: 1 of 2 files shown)


api/src/Page/Sample.php
```diff
@@ -960,6 +997,59 @@ function _update_sub_sample_full()
...
+        $total_updated = 0;
+
+        foreach ($chunks as $chunk) {
+            $id_list = implode(',', $chunk);
+
+            $this->db->pq("UPDATE diffractionplan
+                SET requiredresolution=:1, experimentkind=:2, preferredbeamsizex=:3, preferredbeamsizey=:4,
+                    exposuretime=:5, boxsizex=:6, boxsizey=:7, axisstart=:8, axisrange=:9,
+                    numberofimages=:10, transmission=:11, energy=:12, monochromator=:13
+                WHERE diffractionplanid IN ($id_list)", $set_args);
```
gfrn: Would it be better to do something like in like 663, using placeholders instead of dropping the IDs directly into the query? Just so we avoid adding any possible avenues for SQL injection

client/src/js/modules/imaging/views/queuecontainer.js
```diff
@@ -704,23 +704,36 @@ define(['marionette',
...
-                    Promise.allSettled(promises)
-                    .then(updateButtonAfterProcessing)
-                    .catch(error => {
-                        console.error("Error after promises settled: ", error);
-                        updateButtonAfterProcessing()
-                    })
-                } else {
-                     updateButtonAfterProcessing()
-                }
+                var fields = [
```
gfrn:
```suggestion
                const fields = [
```

gfrn (approved):

PR #1075, opened 2026-06-04
LIMS-2151: Allow summary page to be for visits/proteins/groups

ndg63276:
**JIRA ticket**: [LIMS-2151](https://jira.diamond.ac.uk/browse/LIMS-2151)

**Summary**:

Use the same summary page to compare data processing from an entire protein, or from an entire sample group, as well as from an entire visit.

**Changes**:
- Pass protein id or sample group id to the backend, check they belong to the proposal, add additional joins as needed to get the data
- Use suitable csv filename for each
- Always get proteins in the same order
- Extend sample groups collection into a type that can generate a dropdown list
- Add extra routes and triggers for visit/protein/group summary pages
- Add dropdowns to the summary page to change between visit/protein/group, and to select which visit/protein/group to view
- Add css to make the dropdowns look similar to the text
- Add link to the summary page from the visit/protein/group list pages

**To test**:
- Go to any proposal, then go to /visits, check any visit with data now has a 4th button marked View Summary, that takes you to /dc/summary/visit/<visit>. Check the first dropdown at the top says "Visit", and the second says the correct visit.
- Go to /proteins, check each protein now has a 2nd button marked View Summary, that takes you to /dc/summary/protein/<proteinid>. Check the first dropdown at the top says "Protein", and the second says the correct acronym.
- Go to /samples/groups/, check each group now has a 2nd button marked View Summary, that takes you to /dc/summary/group/<groupid>. Check the first dropdown at the top says "Group", and the second says the correct group name.
- Check you can change groups and the data updates (not all groups have data). Check the "View Sample Groups" and "Download CSV" buttons work.
- Check you can change the type to Visit, and then select a visit, and the data updates. Check the "All Data Collections" and "Download CSV" buttons work.
- Check you can change the type to Protein, and then select a protein, and the data updates. Check the "View Protein" and "Download CSV" buttons work.
- Check the various filters (eg Min Completeness) and the search bar still work.

### DIFF
--- a/client/src/js/templates/dc/summary.html
+++ b/client/src/js/templates/dc/summary.html
@@ -16,14 +16,12 @@
         <a href="/dc/visit/<%-VISIT%>" class="button"><i class="fa fa-search"></i> All Data Collections</a>
         <a href="<%-APIURL%>/processing/summary/visit/<%-VISIT%>/csv/1" class="button csv"><i class="fa fa-file-o"></i> Download CSV</a>
     </div>
-    <% } %>
-    <% if (typeof PROTEINID !== 'undefined' && PROTEINID) { %>
+    <% } else if (typeof PROTEINID !== 'undefined' && PROTEINID) { %>
     <div class="r">
         <a href="/proteins/pid/<%-PROTEINID%>" class="button"><i class="fa fa-search"></i> View Protein</a>
         <a href="<%-APIURL%>/processing/summary/protein/<%-PROTEINID%>/csv/1" class="button csv"><i class="fa fa-file-o"></i> Download CSV</a>
     </div>
-    <% } %>
-    <% if (typeof SAMPLEGROUPID !== 'undefined' && SAMPLEGROUPID) { %>
+    <% } else if (typeof SAMPLEGROUPID !== 'undefined' && SAMPLEGROUPID) { %>
     <div class="r">
         <a href="/samples/groups" class="button"><i class="fa fa-search"></i> View Sample Groups</a>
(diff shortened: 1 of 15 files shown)


client/src/js/templates/dc/summary.html
```diff
@@ -1,10 +1,34 @@
...
 
 <div class="options data_collection">
+    <% if (typeof VISIT !== 'undefined' && VISIT) { %>
     <div class="r">
         <a href="/dc/visit/<%-VISIT%>" class="button"><i class="fa fa-search"></i> All Data Collections</a>
-        <a href="<%-APIURL%>/processing/visit/<%-VISIT%>/csv/1" class="button csv"><i class="fa fa-file-o"></i> Download CSV</a>
+        <a href="<%-APIURL%>/processing/summary/visit/<%-VISIT%>/csv/1" class="button csv"><i class="fa fa-file-o"></i> Download CSV</a>
+    </div>
+    <% } %>
+    <% if (typeof PROTEINID !== 'undefined' && PROTEINID) { %>
```
gfrn: Is this required? If it is undefined or null, it's going to be falsy, so we can probably rewrite this as `if (PROTEINID)` unless there's a caveat I'm not aware of

ndg63276: Without the check for undefined for VISIT/PROTEINID/SAMPLEGROUPID, you just get "VISIT is not defined" errors, though it could be tidied up into else-if clauses.

gfrn (approved):

PR #1076, opened 2026-06-05
LIMS-2208: Show dose on UDC strategy results

ndg63276:
**JIRA ticket**: [LIMS-2208](https://jira.diamond.ac.uk/browse/LIMS-2208)

**Summary**:

Add some more details to the UDC Strategy results

**Changes**:
- Display subwedge dose if non-zero, and hide exposure time if zero
- Show the total rotation as well as per image rotation
- Hide the 'relative transmission' column, only show 'absolute transmission', rename as '(Max) transmission'.
- Add some highlighting to rows to show which wedges belong together

**To test**:
- Go to a data collection with a result from the UDC strategy, eg /dc/visit/nt44198-3/id/23192087
- Open the Strategies bar to display the UDC Strategy results
- Check total rotation, max transmission and dose are displayed
- Check relative transmission and exposure time are hidden
- Check the 'Native' wedges are coloured slightly darker to show they belong together

### DIFF
--- a/client/src/js/templates/dc/dc_strategy.html
+++ b/client/src/js/templates/dc/dc_strategy.html
@@ -2,29 +2,29 @@
 <h2><%-TYPE%></h2>
 
 <%
-    var showUnitCell = CELL && (CELL.SG || CELL.A || CELL.B || CELL.C || CELL.AL || CELL.BE || CELL.GA)
-    var showChi = _.some(STRATS, function(r) {
+    const showUnitCell = CELL && (CELL.SG || CELL.A || CELL.B || CELL.C || CELL.AL || CELL.BE || CELL.GA)
+    const showChi = STRATS.some(function(r) {
         return r.CHI && parseFloat(r.CHI) !== 0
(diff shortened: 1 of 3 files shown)


client/src/js/templates/dc/dc_strategy.html
```diff
@@ -66,15 +72,25 @@ <h2><%-TYPE%></h2>
...
         </tr>
     </thead>
     
-    <% _.each(STRATS, function(r, i) { %>
-    <tr class="stratrow" draggable="true" sid="<%-i%>">
+    <%
+    var highlight = true;
+    var lastId = null;
+
+    _.each(STRATS, function(r, i) {
```
gfrn: Could we rewrite this with `map` or a native function? Just so we don't add any more things using underscore and make updating it harder

ndg63276: Yep, good idea.

gfrn (approved):

PR #1085, opened 2026-07-08
LIMS-2137: Update vee-validate to v3

ndg63276:
**JIRA tickets**: [LIMS-2137](https://jira.diamond.ac.uk/browse/LIMS-2137) and [LIMS-2271](https://jira.diamond.ac.uk/browse/LIMS-2271)

**Summary**:

Update the vee-validate library from v2.x to v3.x.

**Changes**:
- As there were a few vee-validate references in scipion.html, and that is no longer used (and would be in PaTo anyway), that file has been removed
- As there were lots of vee-validate references in the CONEXS submission pages, and @jacobfilik had already asked to remove them, those have been removed
- Without CONEXS submission, we could use the general menu, rather than needing specific menus for "conexs" and "sp" beamline types, so they have been removed
- Update Vee-Validate to ^3.0.0
- Use newer style of imports for main script and custom rules
- Use the `dirty` flag instead of `changed`
- Rewrite custom riles in new style, remove unused `field_exists` rule, add `decimal` rule that was native in v2 but doesnt exist in v3
- Remove references to `this.$validator` and `errors.has()`
- Rewrite to use `<validation-provider rules=` instead of `v-validate=`
- Remove Promise polyfill as it was only needed for IE

**To test**:
- Test sending feedback by going to /feedback (emails may not work on dev but that is not related to this change)
- Go to a b18, i18 or i20 proposal, check there is no menu item for Conexs submission and /conexs gives a 404
- Test creating an MX shipment, an MX container, and samples within the container. Check the fields all display and there are no errors in the console.
- Test creating an XPDF simple sample:
  - Go to an XPDF proposal eg cm33910, then go to Phases from the main menu
  - Click on one of the phases, then click New Simple Sample at the top
  - Give it a unique acronym, fill in the composition, density and packing fraction, and select Borosilicate 1mm OD for the capillary, and click Add Sample
  - Check there are no errors in the console.
- Go back to the Add Simple Sample page, then select File Upload For Multiple Samples. Check this file uploads ok:
[valid_file.csv](https://github.com/user-attachments/files/29804623/valid_file.csv)


gfrn (approved):
Looks good to me, seems like everything works

